PR #135, opened 2015-02-10
adding spaces after # headers in README for proper rendering

mrjf:
required by CommonMark spec; GitHub will render them as headers without spaces, but NPM does not: https://www.npmjs.com/package/proj4

I'm in the process of creating a PR that will add Meteor packaging to proj4js, so you will get a PR for that soon. But I thought regardless of whether you decide you want to support that, you might want these minor changes to the README so the docs will render properly on npmjs.com.

Thank you for proj4js, it's a terrific tool.


calvinmetcalf:
I can publish this tomorrow

mrjf:
Thank you!

Issue #138, opened 2015-03-03
ft and us-ft units in proj4 def are ignored


ahocevar:
See https://github.com/openlayers/ol3/issues/3276#issuecomment-76395726.

fredj:
This issue can be closed, fixed with #139

PR #139, opened 2015-03-03
Add handling for +units in projString

ahocevar:
This adds units constants for ft and us-ft, so the to_meter property can be set properly in the projString parser.


ahocevar:
Fixes #138.

ahocevar:
@calvinmetcalf, any opinion on this?

calvinmetcalf:
sorry +1

PR #140, opened 2015-03-19
Fix WKT handling for stereographic projections on Southern hemisphere

ahocevar:
WKT definitions for stereographic projections on the Southern hemisphere use projection names that are not listed as names in `stere.js`. Also, WKT definitions for such SRSs lack a definition of the standard parallel, which I covered in an extra cleanup step.

### DIFF
--- a/lib/projections/stere.js
+++ b/lib/projections/stere.js
@@ -163,4 +163,4 @@ exports.inverse = function(p) {
   return p;
 
 };
-exports.names = ["stere"];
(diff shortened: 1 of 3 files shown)


ahocevar:
@calvinmetcalf If you have a spare cycle, a quick review would be much appreciated. Thanks!

calvinmetcalf:
Looks good

ahocevar:
Thanks for the review!

Issue #141, opened 2015-03-31
Custom projection - wrong transformation parameters used


imalovic:
When defining custom projection in  Openlayers3 using proj4.defs[], and using 7P to define transformation to WGS84, I  have noticed the rotation parameters and scale are being recalculated multiple times to radians/scale, thus when transforming the wrong values are used.
I have hardcoded rotation parameters and scale for my projection in proj4-src.js, and the transformed coordinates are good now. I have checked the transformed coordinates using FME and ILWIS.

Can someone confirm the same behaviour, or the problem is at my side?

ahocevar:
Can you share a test case or JSFiddle or CRS definition that shows the issue? Along with a transform example and the expected result? Thanks. It would also be good to see your workaround.

JelmerBaas:
I have the exact same issue, and have been debugging for ages now. A simple test-case: http://jsfiddle.net/ajks4jum/1/

This also happens in OL3 - every time I reproject a point in RD, it moves somewhere else. 

The code that's doing this is
    if (proj.datum_params.length > 3) {
      if (proj.datum_params[3] !== 0 || proj.datum_params[4] !== 0 || proj.datum_params[5] !== 0 || proj.datum_params[6] !== 0) {
        this.datum_type = PJD_7PARAM;
        proj.datum_params[3] *= SEC_TO_RAD;
        proj.datum_params[4] *= SEC_TO_RAD;
        proj.datum_params[5] *= SEC_TO_RAD;
        proj.datum_params[6] = (proj.datum_params[6] / 1000000.0) + 1.0;
      }
    }

After which the original definition is screwed up.

imalovic:
OMG,
I have completely forgot about this....Sorry...

The problem is in loop at the lines 744, 745, 746, 747 of proj4-src.js..put a break point there and you will see each time (except the first) you call a transformation, previously recalculated parameters are used...

As I recall we made a quick fix by hardcoding if condition which switches at frist pass, so when it tries again it steps over the code...

JelmerBaas:
I think I actually managed to fix it:

var datum = function(proj) {
  if (!(this instanceof datum)) {
    return new datum(proj);
  }
  this.datum_type = PJD_WGS84; //default setting
  if (!proj) {
    return;
  }
  if (proj.datumCode && proj.datumCode === 'none') {
    this.datum_type = PJD_NODATUM;
  }

  if (proj.datum_params) {
    this.datum_params = proj.datum_params.slice(0);

[...]

PR #155, opened 2015-09-29
Fix issue with custom 7 point projection

JelmerBaas:
This change fixes the issue described at https://github.com/proj4js/proj4js/issues/141 by copying the datum_params object and change the copy, instead of changing the original.

### DIFF
--- a/lib/datum.js
+++ b/lib/datum.js
@@ -20,11 +20,7 @@ var datum = function(proj) {
   }
 
   if (proj.datum_params) {
-    // Copy datum_params object to fix issue with custom 7 point projection (https://github.com/proj4js/proj4js/issues/141)
-    this.datum_params = proj.datum_params.slice(0);
-    for (var i = 0; i < this.datum_params.length; i++) {
-      this.datum_params[i] = parseFloat(this.datum_params[i]);
-    }
+    this.datum_params = proj.datum_params.map(parseFloat); 
     if (this.datum_params[0] !== 0 || this.datum_params[1] !== 0 || this.datum_params[2] !== 0) {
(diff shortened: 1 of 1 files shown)


lib/datum.js
```diff
@@ -18,23 +18,27 @@ var datum = function(proj) {
   if (proj.datumCode && proj.datumCode === 'none') {
     this.datum_type = PJD_NODATUM;
   }
+
   if (proj.datum_params) {
-    for (var i = 0; i < proj.datum_params.length; i++) {
-      proj.datum_params[i] = parseFloat(proj.datum_params[i]);
+    // Copy datum_params object to fix issue with custom 7 point projection (https://github.com/proj4js/proj4js/issues/141)
```
calvinmetcalf: we don't need this comment, comments like these have a tendency to pile up over time to make the code hard to read

lib/datum.js
```diff
@@ -18,23 +18,27 @@ var datum = function(proj) {
...
+    this.datum_params = proj.datum_params.slice(0);
+    for (var i = 0; i < this.datum_params.length; i++) {
+      this.datum_params[i] = parseFloat(this.datum_params[i]);
     }
```
calvinmetcalf: we could combine these 4 lines into 

```js
this.datum_params = proj.datum_params.map(parseFloat); 
```

calvinmetcalf:
it looks good, if your able to make the changes I suggested then thats great, otherwise I can fix them when I merge it in

JelmerBaas:
Took some time to find the edit button, but I've made the changes. Thanks for taking the time to explain.

Issue #156, opened 2015-10-06
to_meter conversion factor wrong for GEOGCS WKT


tbarsballe:
In lib/wkt.js, the `to_meter` conversion factor is set directly from the WKT UNIT (see https://github.com/proj4js/proj4js/blob/master/lib/wkt.js#L114), regardless of the context.

From http://www.geoapi.org/3.0/javadoc/org/opengis/referencing/doc-files/WKT.html#UNIT (emphasis mine):

> UNIT

> This describes units used for values elsewhere within the parent WKT clause (sometimes including descendants of the parent clause). The physical dimension (i.e. type) of the units is determined by context. For example, in a **GEOGCS the type of the units is angular**. In a VERT_CS the type of the units is linear. Within a UNIT clause, the units are described by relating them to a fundamental unit of that type with a conversion factor. For linear units, the conversion factor is the scalar value that converts the described units into meters. **For angular units, the conversion factor is the scalar value that converts the described units into radians**.

`wkt.js` should be updated to take into account the context, using the correct conversion factor of `wkt.UNIT.convert * a` for GEOGCS (Where a is the earth radius as obtained from the DATUM).

calvinmetcalf:
wouldn't this depend on the datum ?

tbarsballe:
Yes, it would, thanks. I have updated the issue description and the PR accordingly.

ahocevar:
Ok, this will give an approximation which is only accurate when measuring east <-> west at the equator, because a is the semi-major axis. But that's fine I think.

PR #157, opened 2015-10-06
Use proper to_meter conversion factor for OGC WKT GEOGCS

tbarsballe:
Fix for https://github.com/proj4js/proj4js/issues/156

6370997 is the normal earth radius, borrowed from https://github.com/openlayers/ol3/blob/master/src/ol/sphere/normal.js#L11
If anyone has a more appropriate earth radius value, feel free to suggest it.

### DIFF
--- a/lib/wkt.js
+++ b/lib/wkt.js
@@ -111,7 +111,13 @@ function cleanWKT(wkt) {
       wkt.units = 'meter';
     }
     if (wkt.UNIT.convert) {
-      wkt.to_meter = parseFloat(wkt.UNIT.convert, 10);
+      if (wkt.type === 'GEOGCS') {
+        if (wkt.DATUM && wkt.DATUM.SPHEROID) {
+          wkt.to_meter = parseFloat(wkt.UNIT.convert, 10)*wkt.DATUM.SPHEROID.a;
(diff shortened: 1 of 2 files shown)


lib/wkt.js
```diff
@@ -111,7 +111,11 @@ function cleanWKT(wkt) {
       wkt.units = 'meter';
     }
     if (wkt.UNIT.convert) {
-      wkt.to_meter = parseFloat(wkt.UNIT.convert, 10);
+      if (wkt.type === 'GEOGCS') {
+        wkt.to_meter = parseFloat(wkt.UNIT.convert, 10)*6370997;
```
ahocevar: You shouldn't be using the constant `6370997` directly here. This value has to be taken from the correct datum.

tbarsballe: Thanks, fix in progress.

tbarsballe: if `wkt.type === 'GEOGCS'` but `DATUM` is undefined, what should the behaviour be?
* use 6370997 ("Normal" sphere)
* throw an error
* do not set to_meter

ahocevar:
Looks good now. Thanks for this improvement.

PR #186, opened 2016-04-07
Fix standalone build and add ortho to default build

### LINKED ISSUE #185: ortho not in default build
Hi, when I attempt to use the ortho projection in 2.3.14, I get an "uncaught" error. I'm attempting to use the same projection as what is mentioned in #170 - that is +proj=ortho +lat_0=90 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs .

I see that ortho was not added to the projs array in Gruntfile.js following the resolution of #170 - is this the reason? If I add 'ortho' to the projs array, then I can generate a custom build without errors. However, I'm apparently missing a step, because any custom build that I create (even using the defaults) results in a proj4-src.js file that, when loaded in a browser, doesn't actually define proj4 (I'm new to building my own js-related stuff).

ahocevar:
Fixes #185.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -56,7 +56,9 @@ module.exports = function(grunt) {
           'dist/proj4-src.js': ['lib/index.js'],
         },
         options: {
-          standalone: 'proj4',
+          browserifyOptions: {
+            standalone: 'proj4'
+          },
           alias: [
             './projs:./includedProjections'
             ]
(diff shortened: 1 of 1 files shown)


probins:
any chance of this being merged so standalone build works properly?

andest01:
Once this gets updated from `2.3.14` then I'd love to do a `npm update`. I just ran into this ortholess problem last night. 👍

calvinmetcalf:
I'm going to let @ahocevar do that as last time I tried I broke stuff

ahocevar:
v2.3.15 published.

PR #188, opened 2016-04-29
Quoteless WKT strings

sainsb:
Came across the proj string at here:
http://library.oregonmetro.gov/rlisdiscovery/lrt_line.zip

this:
`PROJCS[NAD_1983_HARN_StatePlane_Oregon_North_FIPS_3601_Feet_Intl,GEOGCS[GCS_North_American_1983_HARN,DATUM[D_North_American_1983_HARN,SPHEROID[GRS_1980,6378137.0,298.257222101]],PRIMEM[Greenwich,0.0],UNIT[Degree,0.0174532925199433]],PROJECTION[Lambert_Conformal_Conic],PARAMETER[False_Easting,8202099.737532808],PARAMETER[False_Northing,0.0],PARAMETER[Central_Meridian,-120.5],PARAMETER[Standard_Parallel_1,44.33333333333334],PARAMETER[Standard_Parallel_2,46.0],PARAMETER[Latitude_Of_Origin,43.66666666666666],UNIT[Foot,0.3048]]`

Lack of quotes breaks WKT JSON.parse command.
Added a regex to ensure that strings are quoted in the WKT string.
`wkt = wkt.replace(/\[(?=.[a-zA-Z_].*)([a-zA-Z0-9_]+)/g,'["$1"')`

Added a test

### DIFF
--- a/lib/wkt.js
+++ b/lib/wkt.js
@@ -111,7 +111,13 @@ function cleanWKT(wkt) {
       wkt.units = 'meter';
     }
     if (wkt.UNIT.convert) {
-      wkt.to_meter = parseFloat(wkt.UNIT.convert, 10);
+      if (wkt.type === 'GEOGCS') {
+        if (wkt.DATUM && wkt.DATUM.SPHEROID) {
+          wkt.to_meter = parseFloat(wkt.UNIT.convert, 10)*wkt.DATUM.SPHEROID.a;
+        }
+      } else {
+        wkt.to_meter = parseFloat(wkt.UNIT.convert, 10);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


calvinmetcalf:
that file is wrong http://docs.opengeospatial.org/is/12-063r5/12-063r5.html

calvinmetcalf:
or put more diplomatically 😃 that would seem to be a clear case of a malformed proj file and I don't think it's a good idea for proj4 to include support for what seems to be a one off error by a worker at the Portland metro council.  If there is some program or quasi standard somewhere that creates files like this then we can talk more but we support a lot of crazy formats that people intentionally make, we don't need to be supporting formats that people create by accident.

sainsb:
what you say is true.
I'm looking in to what may have generated that prj file... might've been FME in which case we should forward this thread to Safe.

kevinmehall:
We're running into this with WKT generated by [Apache SIS](https://sis.apache.org/). In the document linked above, text is specified to be double-quoted, but there are also enumerations, which are not quoted and are case-insensitive.

Several examples in that spec use unquoted enumerations that look just like the problematic part of our WKT from SIS, e.g:
```
      CS[Cartesian,2],
                 AXIS["northing (X)",north,ORDER[1]],
                 AXIS["easting (Y)",east,ORDER[2]],
                 LENGTHUNIT["German legal metre",1.0000135965]
```

calvinmetcalf:
wkt stuff is now in it's [own repo](https://github.com/proj4js/wkt-parser) so further discussion should go there especially if we have examples of well formed wkt that don't parse

PR #196, opened 2016-07-12
Fix IE8 bug caused by exports.null

andymantell:
IE8 chokes on `exports.null` in `lib/constants/grids.js` because it gets a bit confused by null meaning _null_. The attached fix doesn't change the structure of the object but it avoids the IE8 issue by putting it in quotes.


ahocevar:
This is a good change. Thanks @andymantell.

andymantell:
Thanks for the swift response @ahocevar, very much appreciated!

PR #200, opened 2016-08-23
clean up, remove all the grid shift code that isnt used

calvinmetcalf:
also fix an actual bug where GEOGCS by itself at the begining of a proj string could cause problems


ahocevar:
Awesome, please merge.

PR #201, opened 2016-08-30
attempt at improving the flow of data

calvinmetcalf:
an attempt to be more explicit on what data is being used where and in what functions by explicitly passing parameters instead of mutating things silently


ahocevar:
Thanks for this very beautiful cleanup. Improves readability of the code a great deal, and will make it easier to debug. Please merge.

Issue #203, opened 2016-09-14
Parsing the same CRS with proj4 and WKT results in different parameters


jackielii:
Parsing the same CRS with proj4 and WKT results in different parameters. parameter strings copied from http://epsg.io/4134:

``` javascript
it('should parse wkt and proj4 of the same crs and result in the same params', function() {
  var s1 = 'GEOGCS["PSD93",DATUM["PDO_Survey_Datum_1993",SPHEROID["Clarke 1880 (RGS)",6378249.145,293.465,AUTHORITY["EPSG","7012"]],TOWGS84[-180.624,-225.516,173.919,-0.81,-1.898,8.336,16.7101],AUTHORITY["EPSG","6134"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4134"]]'
  var s2 = '+proj=longlat +ellps=clrk80 +towgs84=-180.624,-225.516,173.919,-0.81,-1.898,8.336,16.7101 +no_defs'
  var crs1 = proj4(s1)
  var crs2 = proj4(s2)
  console.log(crs1.oProj.a)
  console.log(crs2.oProj.a)
  assert.equal(crs1.oProj.a, crs2.oProj.a)
})
```

calvinmetcalf:
it's likely `Clarke 1880 (RGS)` is not getting cleaned up into `clrk80`

PR #204, opened 2016-09-14
add GEOGCS to json in cleanWKT plus test

jackielii:
This fixes the issue #203
But a new issue unrelated to the library is found:

According to epsg.io, 4134<sup>[1](http://epsg.io/4134)</sup> has Ellipsoid Clarke 1880 (RGS)<sup>[2](http://epsg.io/7012-ellipsoid)</sup> which should have inverse flattening 293.465. But proj4js's clrk80<sup>[3](https://github.com/proj4js/proj4js/blob/master/lib/constants/Ellipsoid.js#L73)</sup> <sup>[4](https://github.com/OSGeo/proj.4/blob/master/src/pj_ellps.c#L21)</sup> has 293.4663. 

So is this a proj4's error or a epsg's error?

### DIFF
--- a/lib/wkt.js
+++ b/lib/wkt.js
@@ -92,6 +92,7 @@ function d2r(input) {
 function cleanWKT(wkt) {
   if (wkt.type === 'GEOGCS') {
     wkt.projName = 'longlat';
+    wkt.GEOGCS = wkt;
   }
   else if (wkt.type === 'LOCAL_CS') {
     wkt.projName = 'identity';
--- a/lib/wkt.js
+++ b/lib/wkt.js
@@ -121,15 +121,16 @@ function cleanWKT(wkt) {
     }
   }
 
-  if (wkt.GEOGCS) {
-    //if(wkt.GEOGCS.PRIMEM&&wkt.GEOGCS.PRIMEM.convert){
-    //  wkt.from_greenwich=wkt.GEOGCS.PRIMEM.convert*D2R;
+  var geogcs = wkt.GEOGCS || wkt
+  if (geogcs) {
+    //if(geogcs.PRIMEM&&geogcs.PRIMEM.convert){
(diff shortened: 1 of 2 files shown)


jackielii:
The tests fail because of the inverse flattening parameters issue described above

calvinmetcalf:
there are a couple version of clarke 1880 the one in here seems to be [clarke 1880  (Arc)](http://epsg.io/7013-ellipsoid) as opposed to rgs which you were using

calvinmetcalf:
also it would be great if you could limit pull requests to a single topic, so I'm not sure why you wanted GEOGCS on the wkt object ...

jackielii:
@calvinmetcalf the wkt here already the spheroid.a, but it fails to go into the if branch at line wkt.js:125.

Should it be getting the `a` parameter from the wkt or from the ellipsoid constants?

calvinmetcalf:
the reverse flatning is the rf param not a, a is the major axis and it gets it from the ellipsoid

jackielii:
The original problem is if we don't add the `GEOGCS` property to wkt, it won't have any parameters, then it'll get WGS84's parameters

jackielii:
This is a very naive fix, probably needs to be re-thought.

We can keep the test, as the test should definitely pass

calvinmetcalf:
> We can keep the test, as the test should definitely pass

there are actually multiple datums refereed to as Clarke 1880 so it's unclear that the RGS should be the default one

jackielii:
According to EPSG: http://epsg.io/4134, it should be Clarke 1880 (RGS): http://epsg.io/7012-ellipsoid

jackielii:
Could you take another look at the fix.

Without this fix, all the wkt that's geodetic datum will fail to get any parameters, then it'll fall into deriveConstants.js:33 which will assign WGS84's sphere parameters to it.

calvinmetcalf:
ok I see what your doing now

PR #205, opened 2016-09-19
Added Typescript definition to DT

DenisCarriere:
Improvements of this typescript definition are welcomed.

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/proj4

To install

```bash
$ typings install --save --global dt~proj4
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,16 @@ TypeScript implementation was added to the [DefinitelyTyped repository](https://
 
 **Install definition**
 
+Using DefinitelyTyped
+
 ```bash
 $ typings install --save --global dt~proj4
+```
+
+Using Typescript 2.0
+
+```bash
+$ npm install --save @types/proj4
 ```
 
 ## Developing
--- a/README.md
+++ b/README.md
@@ -114,6 +114,16 @@ proj4.defs can also be used to define a named alias:
 proj4.defs('urn:x-ogc:def:crs:EPSG:4326', proj4.defs('EPSG:4326'));
 ```
 
+## TypeScript
+
+TypeScript implementation was added to the [DefinitelyTyped repository](https://github.com/DefinitelyTyped/DefinitelyTyped).
+
+**Install definition**
+
+```bash
+$ typings install --save --global dt~proj4
+```
+
 ## Developing
 to set up build tools make sure you have node and grunt-cli installed and then run `npm install`
 
--- a/README.md
+++ b/README.md
@@ -118,16 +118,6 @@ proj4.defs('urn:x-ogc:def:crs:EPSG:4326', proj4.defs('EPSG:4326'));
 
 TypeScript implementation was added to the [DefinitelyTyped repository](https://github.com/DefinitelyTyped/DefinitelyTyped).
 
-**Install definition**
-
-Using DefinitelyTyped
-
-```bash
-$ typings install --save --global dt~proj4
-```
-
-Using Typescript 2.0
-
 ```bash
 $ npm install --save @types/proj4
 ```


calvinmetcalf:
I thought typescript was moving off of this system?

I let me do a quick look the the file to see if I notice anything

calvinmetcalf:
```ts
  interface InterfaceCoordinates {
        x: number,
        y: number,
        z?: number,
        m?: number
}
```

we don't do anything with measure, I think it would actually get stripped off if you actually tried to do anything

```ts
   interface InterfaceProjection {
        title: string
        projName: string
        ellps: string
        datumCode: string
        units: string
        a: number
        rf: number
        ellipseName: string
        b: number
        a2: number
        b2: number
        es: number
        e: number
        ep2: number
        k0: number
        axis: string
        datum: InterfaceDatum
        init: typeof proj4.Proj,
        forward(coordinates: typeof TemplateCoordinates): Array<number>
        inverse(coordinates: typeof TemplateCoordinates): Array<number>
        names: Array<string>
        to_meter(value: number): any
        from_greenwich(value: number): any
}
```

this is totally variable, subject to change and all except forward, and inverse should probably be considered an implementation detail

DenisCarriere:
@calvinmetcalf What is Typescript moving towards?? DefinitelyTyped has 8300+ stars on Github, 21,000 commits & 2500 contributors, I haven't seen anything that's more popular then that.

There's some large projects that are being maintained on DT, one I particular like is `Lodash`.

Thanks for your input about the Definition, I'll make those modifications.

It make sense to remove the extra variables in the `InterfaceProjection`.

I've found the `x,y,z,m` coordinates in your "new" `toPoint` method.

https://github.com/proj4js/proj4js/blob/master/lib/common/toPoint.js

```javascript
module.exports = function (array){
  var out = {
    x: array[0],
    y: array[1]
  };
  if (array.length>2) {
    out.z = array[2];
  }
  if (array.length>3) {
    out.m = array[3];
  }
  return out;
};
```

calvinmetcalf:
Oh I have no idea I just thought I heard that they were moving away from that towards some sort of npm based thing? I really have no experience with typed script I may have misunderstood the announcement.

Oh hey forgot about that, yeah guess it doesn't hurt to leave m in

DenisCarriere:
Typescript might of announced something, however it hasn't been picked up by the community, so far all the popular libraries are being hosted on DT.

Some of the definitions on npm are very outdated compared to DT.

DenisCarriere:
The new way to install typings will be done using npm in Typescript 2.0

```bash
$ npm install --save @types/lodash
```

Definitely Typed will still remain the place to author definitions.

More info about this: https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

calvinmetcalf:
Ah that's what I was thinking of

DenisCarriere:
@calvinmetcalf Essentially the new Typescript definitions will be hosted by DefinitivelyTyped.

A reply from one of the DT collaborators:

> We have to do manual porting of the latest changes to master to the types-2.0 branch. This should happen about once a week.

When this happens, you will be able to install the new definition using.

```bash
$ npm install @types/proj4
```

All Types will be hosted on npm.

https://www.npmjs.com/~types

calvinmetcalf:
should we add a note about that here as well or is it too early for that?

DenisCarriere:
@calvinmetcalf That could be possible, this install only works once Typescript 2.0 is released.

DenisCarriere:
@calvinmetcalf Any news on this PR?

Issue #208, opened 2016-09-22
Issues with webpack


atlefren:
Hi!

Sorry if I'm reporting a known bug, but I couldn't find any reference to the problem I'm facing

My problem is this

I've installed proj4 2.3.15 using npm and try to include it in my project using webpack:

`import proj4 from 'proj4';`

This works fine, but webpack gives me an error

```
WARNING in ./~/proj4/dist/proj4.js
Critical dependencies:
1:407-414 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/proj4/dist/proj4.js 1:407-414
```

Well, I guess I'll do something about that is my first thought, but then I check the package.json file for proj4, and main is set to lib/index.js, and all seems fine. The only reference to dist/proj4.js I can find is in the "jam"-block. Stange!

I then tried to remove the yam-block, thinking that this config might be preferred over main (for some, to me, unknown reason, but it seems like @calvinmetcalf  is aware of this: https://github.com/webpack/webpack/issues/1118)

So, removing the yam-block I get progress, sort of:

```
ERROR in ./~/proj4/package.json
Module parse failed: /home/atlefren/path/to/project/node_modules/proj4/package.json Unexpected token (2:9)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token (2:9)
[snip]
 @ ./~/proj4/lib/index.js 10:16-42
```

lib/index.js, line 10 is this

`proj4.version = require('../package.json').version;`

Eureka!

I presume that node has no problems reading json-files, but webpack does not support reading json files out of the box. By installing json-loader and following the advice here: https://github.com/webpack/webpack/issues/965#issuecomment-97186184 I manage to import proj4 using webpack, but this is a bit of a hassle.

Unfortunately I cannot think of a way to solve the issue, so unless someone solves the problem I guess the project is better of with the jam-config beeing used, although this is not ideal.

Would be happy to test fixes if anyone has ideas!

Current env:
node: v6.4.0
npm: 3.10.3
webpack: 1.13.2

calvinmetcalf:
so the jam thing is a bug in webpack, we could probably remove it because I don't really feel like jam is going anywhere anymore, as for the json thing, yes that is something that webpack does that hits *many* projects so you I'd recommend always having the json loader because you if your loading modules from npm everyone assumes your going to have it and it's really really unhelpful on the part of webpack not to include it by default.

[...]

PR #209, opened 2016-09-23
Remove jam config

atlefren:
As explained in   #208 the Jam-config is used by webpack when requireing/importing proj4. This means that the minifed version is imported, which triggers a webpack-warning.

Jam seems to be discontinued (https://github.com/caolan/jam), so the easiest solution to this problem is to remove the jam-config

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ Depending on your preferences
 ```bash
 npm install proj4
 bower install proj4
-jam install proj4
 component install proj4js/proj4js
 ```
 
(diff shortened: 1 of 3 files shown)


calvinmetcalf:
ok there were a few more jam things but I was able to use the nifty new ability to push to pull requests to fix them, will merge if everything passes

atlefren:
Ah, sorry about that ;) "push to pull request" was new to me, nice !

PR #210, opened 2016-09-23
Update readme to reflect webpack issues

atlefren:
Mention the webpack issues (and workarounds) described in #208 in readme.

This makes sense once #209 is merged in


calvinmetcalf:
actually we can fix this in a different way

atlefren:
If the other way means less hassle for users of the package I'm all for it! Let me know if you need some help

calvinmetcalf:
see #211

PR #211, opened 2016-09-23
beter versioning

calvinmetcalf:
alternative way to do #210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ coverage
 projs.js
 .DS_STORE
 dist
+lib/version-browser.js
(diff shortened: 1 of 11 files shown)


calvinmetcalf:
the downside is we have a dependency, we could also statically generate lib/version-browser durring the publish step

calvinmetcalf:
ok this one instead has a placeholder file which gets overwritten when we build things before publishing

atlefren:
This approach sounds much better than requireing the end-user to install packages and tweak configs!

calvinmetcalf:
@atlefren fyi if you don't want to install configs/packages you should probably avoid webpack, it's designed with app builders installing global transforms in mind.  But this particular fix should also help browserify users as well by making their build smaller.

calvinmetcalf:
@ahocevar I actually did it without the extra dep so even better

ahocevar:
👍 thanks @calvinmetcalf.

PR #221, opened 2016-12-16
WIP: use a seperate wkt parser

calvinmetcalf:
we've had a lot of issues with wkt files, this pulls out the wkt handling and uses a real parser instead of the regexes and the json.parse


calvinmetcalf:
subsumed into #224

Issue #187, opened 2016-04-22
Instability for UTM projection


perliedman:
When repeatedly projecting and unprojecting a coordinate in UTM projection, proj4 seems to be surprisingly unstable on northern latitudes, about an order of magnitude more unstable than `cs2cs`.

I've run the following test program:

``` javascript
var proj4 = require('proj4'),
    proj = proj4(proj4.WGS84, process.argv[2]),
    ll = [31, 70],
    p0 = proj.forward(ll);

for (var i = 0; i < 10; i++) {
    p = proj.forward(ll);
    ll = proj.inverse(p);
}

var dx = p[0] - p0[0],
    dy = p[1] - p0[1];

console.log(Math.sqrt(dx*dx + dy*dy));
```

Running this with EPSG:32633 (`+proj=utm +zone=33 +datum=WGS84 +units=m +no_defs`), results in a difference of 364 meters.

Doing the same thing with cs2cs (https://gist.github.com/perliedman/44ea458607fba4f121314628c9cee978) gives a difference of 23 meters.

Is this difference expected? It causes trouble for some Leaflet users (https://github.com/Leaflet/Leaflet/issues/4480), since Leaflet internally will sometimes need to project and unproject the same coordinate.

mbloch:
The greater stability of cs2cs is likely because it uses a more numerically stable implementation of transverse mercator for UTM than proj4js ("etmerc" instead of "tmerc"). Perhaps etmerc should be ported to proj4js.

calvinmetcalf:
etmerc instead of tmerc would be very welcome, if anyone wants to open a pull

theashyster:
@perliedman I am now looking more into this and will try to port etmerc to proj4js.
Now I am trying to get as much test data as possible and used your script.

> Doing the same thing with cs2cs (https://gist.github.com/perliedman/44ea458607fba4f121314628c9cee978) gives a difference of 23 meters.

I tried it an it gives me a difference of 0 meters, could this be correct? Which version of proj.4 did you use for your test back on 22 Apr 2016? Can you do a retest of this using the master code of proj.4 lib?

@mbloch I also tried using your https://github.com/mbloch/mapshaper-proj lib and using the mcs2cs tool and for me it looks like the results are more the same as using the cs2cs from proj.4 lib. I see that you have etmerc already ported to js in your lib, so I should probably be able to port it to proj4js from your lib, right? Any hints would help a ton.

perliedman:
[...]

PR #223, opened 2017-01-19
Fix tmerc / utm projection calculations

theashyster:
Reason of doing this is because calculation results for `proj4js` and `cs2cs` differed by a large margin and it meant that there is an issue in the `tmerc` projection calculations.

More details here https://github.com/proj4js/proj4js/issues/187

Hugely inspired by `mapshaper-proj` and how `tmerc` projection is calculated there https://github.com/mbloch/mapshaper-proj/blob/master/src/projections/tmerc.js

- Added common function to adjust zone, called adjust_zone
- Fixed `tmerc` / `utm` projection calculations that were not accurate enough
- Adjusted test data for the more precise calculations

### DIFF
--- a/lib/common/adjust_zone.js
+++ b/lib/common/adjust_zone.js
@@ -13,7 +13,7 @@ module.exports = function(zone, lon) {
   }
   else {
     if (zone > 0 && zone <= 60) {
-      return --zone;
+      return zone - 1;
     }
   }
 };
--- a/lib/common/adjust_zone.js
+++ b/lib/common/adjust_zone.js
@@ -1,0 +1,19 @@
+var adjust_lon = require('./adjust_lon');
+
+module.exports = function(zone, lon) {
+  if (!zone) {
+    zone = Math.floor((adjust_lon(lon) + Math.PI) * 30 / Math.PI);
+
+    if (zone < 0) {
+      return 0;
+    }
+    else if (zone >= 60) {
+      return 59;
+    }
+  }
+  else {
+    if (zone > 0 && zone <= 60) {
+      return --zone;
+    }
+  }
+};
--- a/lib/projections/tmerc.js
+++ b/lib/projections/tmerc.js
@@ -13,9 +13,9 @@ exports.init = function() {
 };
 
 /**
-     Transverse Mercator Forward  - long/lat to x/y
-     long/lat in radians
-   */
+    Transverse Mercator Forward  - long/lat to x/y
+    long/lat in radians
+  */
 exports.forward = function(p) {
   var lon = p.x;
   var lat = p.y;
@@ -73,8 +73,8 @@ exports.forward = function(p) {
 };
 
 /**
-     Transverse Mercator Inverse  -  x/y to long/lat
-   */
+    Transverse Mercator Inverse  -  x/y to long/lat
+  */
 exports.inverse = function(p) {
   var con, phi;
   var lat, lon;
--- a/lib/projections/tmerc.js
+++ b/lib/projections/tmerc.js
@@ -10,6 +10,11 @@ var EPSLN = 1.0e-10;
 var sign = require('../common/sign');
 
 exports.init = function() {
+  this.x0 = this.x0 !== undefined ? this.x0 : 0;
+  this.y0 = this.y0 !== undefined ? this.y0 : 0;
+  this.long0 = this.long0 !== undefined ? this.long0 : 0;
+  this.lat0 = this.lat0 !== undefined ? this.lat0 : 0;
+
   if (this.es) {
     this.en = pj_enfn(this.es);
     this.ml0 = pj_mlfn(this.lat0, Math.sin(this.lat0), Math.cos(this.lat0), this.en);
@@ -30,7 +35,7 @@ exports.forward = function(p) {
   var sin_phi = Math.sin(lat);
   var cos_phi = Math.cos(lat);
 
-  if (this.sphere) {
+  if (!this.es) {
     var b = cos_phi * Math.sin(delta_lon);
 
     if ((Math.abs(Math.abs(b) - 1)) < EPSLN) {
@@ -101,7 +106,7 @@ exports.inverse = function(p) {
   var x = (p.x - this.x0) * (1 / this.a);
   var y = (p.y - this.y0) * (1 / this.a);
 
-  if (this.sphere) {
+  if (!this.es) {
     var f = Math.exp(x / this.k0);
     var g = 0.5 * (f - 1 / f);
     var temp = this.lat0 + y / this.k0;
--- a/lib/projections/tmerc.js
+++ b/lib/projections/tmerc.js
@@ -1,3 +1,6 @@
+// Heavily based on this tmerc projection implementation
+// https://github.com/mbloch/mapshaper-proj/blob/master/src/projections/tmerc.js
+
 var pj_enfn = require('../common/pj_enfn');
 var pj_mlfn = require('../common/pj_mlfn');
 var pj_inv_mlfn = require('../common/pj_inv_mlfn');
--- a/lib/projections/tmerc.js
+++ b/lib/projections/tmerc.js
@@ -1,26 +1,19 @@
-var e0fn = require('../common/e0fn');
-var e1fn = require('../common/e1fn');
-var e2fn = require('../common/e2fn');
-var e3fn = require('../common/e3fn');
-var mlfn = require('../common/mlfn');
+var pj_enfn = require('../common/pj_enfn');
+var pj_mlfn = require('../common/pj_mlfn');
+var pj_inv_mlfn = require('../common/pj_inv_mlfn');
 var adjust_lon = require('../common/adjust_lon');
-var HALF_PI = Math.PI/2;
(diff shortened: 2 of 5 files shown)


calvinmetcalf:
there is a bit of noise in this that it would be nice if you could remove, I'll note them in the code

lib/projections/tmerc.js
```diff
@@ -1,26 +1,19 @@
...
-var e1fn = require('../common/e1fn');
-var e2fn = require('../common/e2fn');
-var e3fn = require('../common/e3fn');
-var mlfn = require('../common/mlfn');
+var pj_enfn = require('../common/pj_enfn');
+var pj_mlfn = require('../common/pj_mlfn');
+var pj_inv_mlfn = require('../common/pj_inv_mlfn');
 var adjust_lon = require('../common/adjust_lon');
-var HALF_PI = Math.PI/2;
-var EPSLN = 1.0e-10;
```
calvinmetcalf: don't move these please

lib/projections/tmerc.js
```diff
@@ -1,26 +1,19 @@
...
-  this.ml0 = this.a * mlfn(this.e0, this.e1, this.e2, this.e3, this.lat0);
+  this.en = pj_enfn(this.es);
+  this.ml0 = pj_mlfn(this.lat0, Math.sin(this.lat0), Math.cos(this.lat0), this.en);
 };
 
-/**
-    Transverse Mercator Forward  - long/lat to x/y
-    long/lat in radians
-  */
+// Transverse Mercator Forward - long/lat to x/y (long/lat in radians)
```
calvinmetcalf: comment style was changed

lib/projections/tmerc.js
```diff
@@ -49,29 +42,37 @@ exports.forward = function(p) {
...
+    var ts = Math.pow(t, 2);
     con = 1 - this.es * Math.pow(sin_phi, 2);
-    var n = this.a / Math.sqrt(con);
-    var ml = this.a * mlfn(this.e0, this.e1, this.e2, this.e3, lat);
+    al = al / Math.sqrt(con);
+    var ml = pj_mlfn(lat, sin_phi, cos_phi, this.en);
 
-    x = this.k0 * n * al * (1 + als / 6 * (1 - t + c + als / 20 * (5 - 18 * t + Math.pow(t, 2) + 72 * c - 58 * this.ep2))) + this.x0;
-    y = this.k0 * (ml - this.ml0 + n * tq * (als * (0.5 + als / 24 * (5 - t + 9 * c + 4 * Math.pow(c, 2) + als / 30 * (61 - 58 * t + Math.pow(t, 2) + 600 * c - 330 * this.ep2))))) + this.y0;
+    x = this.a * (this.k0 * al * (1 +
```
calvinmetcalf: that being said, I like the line breaks here, much more readable

lib/projections/tmerc.js
```diff
@@ -91,45 +92,46 @@ exports.inverse = function(p) {
       lon = adjust_lon(Math.atan2(g, h) + this.long0);
     }
   }
-  else { // ellipsoidal form
```
calvinmetcalf: we probably still want the comment that this is for ellipsoids

lib/common/adjust_zone.js
```diff
@@ -0,0 +1,19 @@
...
+    if (zone < 0) {
+      return 0;
+    }
+    else if (zone >= 60) {
+      return 59;
+    }
+  }
+  else {
+    if (zone > 0 && zone <= 60) {
+      return --zone;
```
calvinmetcalf: `return zone - 1;` woudl be much clearer

calvinmetcalf:
how is this going to work for Norway and its special snowflake zones?

theashyster:
@calvinmetcalf Okay, sorry about the noise, will remove, thanks. Hmm, what do you mean by special snowflake zones?

calvinmetcalf:
utm-zones

theashyster:
@calvinmetcalf Have removed the additional noise in the pull request. Regarding your question, it is hard for me to answer it. Do you have a test case for this? I actually work daily with maps of Norway and we use `utm` zone 33 to cover all Norway and using this `tmerc` implementation improves the current results a lot, but I think that was not your concern, right? Are you concerned that the `adjust_zone` function will detect that the actual zone 32V is calculated to be 31V?

calvinmetcalf:
> and we use utm zone 33 to cover all Norway

do you mean 32 or do you actually use 33?

right so if you have some data in the part of 32v that should be 31v or the part of 33x that should be in the non existent 34x will this lead to incorrect results, for this we might need to compare an outside source like something published by Norway (or a coord from one of your maps)

theashyster:
@calvinmetcalf We actually use 33. This is the `proj4js` string we use `+proj=utm +zone=33 +ellps=GRS80 +units=m +no_defs`.

I am not really sure if that would lead to incorrect results actually. I see, I don't know where to find something published by Norway, but I can try to provide some data from our maps if that can help.

Also, I just thought about that the `adjust_zone` will also impact the `tmerc` code for spheres, but I did not look into that since I did not have any test data for spheres using `tmerc`. But that part should also be updated because of the changes in `utm` regarding zones, right?

theashyster:
I just updated the `tmerc` code for sphere handling, but I still lack of tests for these cases. Had to make a few minor tweaks there. But I would love to write some tests for these changes, but I would need some test data to check upon.

calvinmetcalf:
for test data you probably want to try projecting things in both cs2cs and proj4 (the c version) and maybe another independent one like proj4.j or even esri if you have access, that way you'd be able to be confident that your answer is the answer people expect

theashyster:
Hi @calvinmetcalf 

Yes, I have done that, I used this pull request version of `proj4js`, `cs2cs`, `proj4` (c version) and `mapshaper-proj` to run through the same 10 tests and output the results. Here are the results https://docs.google.com/spreadsheets/d/1k01bfcSE00EmCE9zgeQEem2qZQA38rtyJySZdrBVcow/edit?usp=sharing and they are pretty much identical in all tests that I made.

What is `proj4.j`, I cannot find that one. And what exactly you mean by esri? Which lib? I can try to test with these also.

Anyhow, looks like with the code I have added, the results we get from `proj4js` are more in line with the results we can get from `proj.4` and `mapshaper-proj` libraries.

test/testData.js
```diff
@@ -308,6 +308,80 @@ var testPoints = [
...
+  },
+  {
+    code: '+proj=utm +zone=33 +datum=WGS84 +units=m +no_defs',
+    ll: [31, 70],
+    xy: [1104629.4280255223, 7845845.076400871],
+    acc: {
+      ll: 4
+    }
+  },
+  // these test cases are for Norway snow flake zones
```
calvinmetcalf: 🥇 comment

calvinmetcalf:
hey that looks good, by ESRI I just meant like arcmap if you had access to it, It looks overall pretty good, let me make a last look through first.

calvinmetcalf:
you know what, I'm pretty sure none of the major libraries actually handle UTM over Norway 'correctly' mainly because just knowing the zone number with the the zone letter is technically not enough information to transform correctly, but at this point doing it 'correctly' is probably the wrong choice here compared to doing what everyone else does.

theashyster:
@calvinmetcalf Thanks :)
Unfortunately, I don't have access to `arcmap`. Heh, for sure, take the time to look through it.

One thing I wanted to tell is that I saw that in these snow flakes zones the actual transforms get really unprecise really fast if you choose the wrong zone in the `proj4` string. But it looks like it acts the same way in all of these libraries and that was just the problem with `tmerc`. For `proj.4` in these tests I used version 4.9.2, because it still uses `tmerc` to be able to compare the results, also tested it with `etmerc` and the precision is much more better there and that will be my next pull request.

About the correctness of UTM over Norway, for me it looks like `proj.4` version 4.9.3 with `etmerc` does a really better job at this if we compare to `tmerc`, so I am looking forward on adding it to `proj4js`.

calvinmetcalf:
I did some further research and the latitude bands are actually only part of MGRS not UTM, so not actually relevent

theashyster:
I see, cool. Awesome that this is merged now :)

calvinmetcalf:
yup, haven't published yet tho

theashyster:
:) Thanks for the help.
I will try to make the `etmerc` pull request tomorrow, but that should probably be in a separate release.

calvinmetcalf:
I mean I probably won't get to it today, but I'll probably get to it before the next one gets merged in

calvinmetcalf:
and published

theashyster:
:) Hehe, niiiiiiiice, thanks.

mbloch:
I see what looks like a bug in the `utm.js` file, in the following lines:
```
  var zone = adjust_zone(this.zone, this.long0);
  if (!zone) {
    return;
  }
```
If you initialize the projection using zone 1, the `adjust_zone()` function returns 0, and the initialization fails.

theashyster:
@mbloch You are correct, sadly. I had to check for `undefined` there. True that, will fix it now, but I guess the version cannot be rolled back, right?

PR #224, opened 2017-01-22
Use ES6 modules, use rollup for module bundling

simon04:
Overall, this reduces the size of dist/proj4.js from 73001 to 63914 bytes.


lib/datumUtils.js
```diff
@@ -3,21 +3,21 @@ var PJD_3PARAM = 1;
...
     return false; // false, datums are not equal
-  } else if (source.a !== dest.a || Math.abs(this.es - dest.es) > 0.000000000050) {
-    // the tolerence for es is to ensure that GRS80 and WGS84
+  } else if (source.a !== dest.a || Math.abs(source.es - dest.es) > 0.000000000050) {
+    // the tolerance for es is to ensure that GRS80 and WGS84
     // are considered identical
     return false;
   } else if (source.datum_type === PJD_3PARAM) {
-    return (this.datum_params[0] === dest.datum_params[0] && source.datum_params[1] === dest.datum_params[1] && source.datum_params[2] === dest.datum_params[2]);
+    return (source.datum_params[0] === dest.datum_params[0] && source.datum_params[1] === dest.datum_params[1] && source.datum_params[2] === dest.datum_params[2]);
```
calvinmetcalf: this is a different change

lib/version.js
```diff
@@ -1 +1 @@
-module.exports = require('../package.json').version;
+export default require('../package.json').version;
```
calvinmetcalf: this is almost certainly the wrong way to do this

package.json
```diff
@@ -20,12 +21,14 @@
...
     "grunt-contrib-connect": "~0.6.0",
-    "grunt-contrib-jshint": "~0.8.0",
+    "grunt-contrib-jshint": "~1.1.0",
     "chai": "~1.8.1",
     "mocha": "~1.17.1",
     "grunt-mocha-phantomjs": "~0.4.0",
-    "browserify": "~12.0.1",
-    "grunt-browserify": "~4.0.1",
+    "rollup": "^0.41.4",
+    "rollup-plugin-commonjs": "^7.0.0",
```
calvinmetcalf: we likely want to avoid this dependency and instead also change `mgrs as well`

package.json
```diff
@@ -20,12 +21,14 @@
...
+    "rollup-plugin-node-resolve": "^2.0.0",
```
calvinmetcalf: does this handle the 

```json
   "browser": {
     "./lib/version.js": "./lib/version-browser.js"
   },
```

stuff correctly?

.jshintrc
```diff
@@ -1,4 +1,5 @@
 {
+  "esversion": 6,
```
calvinmetcalf: I think this adds all es6 features not just modules

lib/index.js
```diff
@@ -1,12 +1,21 @@
...
-require('./includedProjections')(proj4);
-module.exports = proj4;
+proj4.Point = Point;
+proj4.toPoint = common;
+proj4.defs = defs;
+proj4.transform = transform;
+proj4.mgrs = mgrs;
+proj4.version = version;
+includedProjections(proj4);
+export default proj4;
```
calvinmetcalf: doing it like this means that anyone importing proj4 looses all the of the benefits of it, of course not doing this breaks compatibility

.jshintrc
simon04: I don't see any other option. Using [`module`](http://jshint.com/docs/options/#module) fails with `The 'module' option is only available when linting ECMAScript 6 code.`

lib/datumUtils.js
simon04: `this` change is intended. What would `this.datum_params` and `this.es` be referring to? Since the function `compareDatums` does not define a class. In the old code are comparing `source.datum_params[1] === dest.datum_params[1]`, `source.datum_params[2] === dest.datum_params[2]`, but not `source.datum_params[0] === dest.datum_params[0]`, contradicting symmetry.

calvinmetcalf:
> Overall, this reduces the size of dist/proj4.js from 73001 to 63914 bytes.

what is the gziped savings ?

simon04:
First of all thank you for your review!

`proj4.js.gz`: 22905 → 21663 (using gzip 1.8).

Concerning custom builds: I'll look into those. In the long run, using ES6 modules w/ tree shaking could completely replace those. A sketch how it could look like:

```javascript
import {proj} from 'proj4/core';
import {proj4_defs} from 'proj4/defs';
import {longlat} from 'proj4/projetions';
proj4_defs('EPSG:4326', '+title=WGS 84 (long/lat) +proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees', longlat);
```

calvinmetcalf:
by my checking dist/proj4.js is only 28,178 so this would actually make it bigger ...

calvinmetcalf:
the projection objects should be objects that we default export, the result of a import * as foo from 'bar' isn't a real object but we do treat ours that way.

calvinmetcalf:
> Concerning custom builds: I'll look into those. In the long run, using ES6 modules w/ tree shaking could completely replace those. 

in theory yes but not without revamping the api and everybody to be using build tools that handle treeshaking so we need to port that over.

simon04:
I just did `npm install proj4` (proj4@^2.3.17) and obtained:
```
$ ls -l node_modules/proj4/dist/         
total 228
-rw-r--r-- 1 simon users  73322 2017-01-24 15:07 proj4.js
-rw-r--r-- 1 simon users 156029 2017-01-24 15:07 proj4-src.js
```

I don't know how you obtained such a small `dist/proj4.js`.

lib/datumUtils.js
calvinmetcalf: that is a good call and I'm actually going to get that in a separate patch to get it in without waiting for this one

calvinmetcalf: c4ddd2576325a950941180809b23235fb4abbc30

.jshintrc
calvinmetcalf: I may just take this opportunity to switch over to eslint

calvinmetcalf:
nope you are right was using the wrong build method so uglified and gzipped we have

before: 22955
after: 21759

or ~5% savings

package.json
simon04: Actually this can be dropped completely, see d940807db509a284756279f9c61b03db54f48542.

calvinmetcalf:
Does this effectively tree shake the rest of the package. Json out?

simon04:
It does, see its [documentation](https://github.com/rollup/rollup-plugin-json#rollup-plugin-json): The `dist/proj4-src.js` only contains
```javascript
var version = "2.3.18-alpha";
// other stuff
proj4$1.version = version;
```

simon04:
In my opinion, we should keep https://github.com/proj4js/proj4js/pull/224#discussion_r97308736, https://github.com/proj4js/proj4js/pull/224#issuecomment-274940707, https://github.com/proj4js/proj4js/pull/224#issuecomment-274941809 for later.

We now got the minimal changes to get ES6 modules together with rollup working. Many other ES6 aspects, refactorings, … can be investigated/applied later.

simon04:
Yay, the builds are passing! :smile:

calvinmetcalf:
I mean for only 5% I'd prefer not to regess functionality, the custom build
stuff should be easy and I can help, the switching to default exports
should actually prevent some bugs and make it faster.

Also this needs to be rebased as I just landed another pull but I can help
with that.

calvinmetcalf:
ok pushed up a fix for the proj exports

calvinmetcalf:
it looks all good, though I need to publish the mgrs changes first

calvinmetcalf:
ok published and added the mgrs dep, I'm going to wait for #226 to land, then back port it into here before merging

calvinmetcalf:
i merged #226 into this one so I can land them both in at once

calvinmetcalf:
heck while we're at it may as well throw in #221

calvinmetcalf:
I suppose our 2 deps (mgrs and wkt-parser) should probably be moved over to dev-dependencies except that if somebody else required it for rollup then they wouldn't get the deps, any thoughts @Rich-Harris ?

simon04:
I would assume that the ES6/webpack/rollup usage increases. So earlier than later there will be problems when dependencies are not specified correctly. For non-ES6/webpack/rollup users the two small dependencies will not make a big difference for download time, and no difference for build size. +1 for keeping them in `dependencies`.

calvinmetcalf:
more browserify and node people getting extraneous npm stuff which they may dislike but wouldn't be a big deal, 

@ahocevar I'd like a review by you before I merge this in as it is a fairly large change, should be minimal changes except for the related etmerc stuff so probably a minor version bump

calvinmetcalf:
also @ahocevar I tried to squash this into 3 commits but I just made a mess so I'll probably use the big green button to squash it into 1 when I merge

simon04:
Squashing 36 commits into 1 commit will lose a lot of history+information how we got there, mixing a few rather unrelated changes into one monster. I'd prefer a (non-fast-forward) merge.

calvinmetcalf:
merged and published as 2.3.18-alpha with the alpha tag on npm so you have to do npm install proj4@alpha to get it

theashyster:
@calvinmetcalf Tried to install it and got an error.
```
>npm install proj4@alpha
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install" "proj4@alpha"
npm ERR! node v4.4.4
npm ERR! npm  v2.15.1
npm ERR! code ETARGET

npm ERR! notarget No compatible version found: proj4@'alpha'
npm ERR! notarget Valid install targets:
npm ERR! notarget ["1.4.0-dev.1","1.4.0-dev.2","1.4.0-dev.3","1.4.0","1.4.1-dev.1","1.4.1","1.5.0-dev.2","1.5.0-dev.3","1.5.0-dev.4","1.5.0-dev.5","2.0.0-dev.1","2.0.0-dev.2","2.0.0-dev.3","2.0.0","2.0.1","2.0.2","2.0.3","2.1.0","2.1.2","2.1.3","2.1.4","2.2.0","2.
2.1","2.2.2","2.3.3","2.3.5","2.3.6","2.3.7","2.3.8","2.3.9","2.3.10","2.3.11","2.3.12","2.3.13","2.3.14","2.3.15","2.3.16","2.3.17"]
```

calvinmetcalf:
```bash
$ npm install proj4@alpha

└─┬ proj4@2.3.18-alpha
  ├── mgrs@1.0.0
  └── wkt-parser@1.1.2
```

maybe try `npm cache clean`

theashyster:
Does not help, could it be that my `npm` version is too old and it does not pick up the `alpha` release? Because if you look at the valid install targets, it does not show up there either.

calvinmetcalf:
no tags have been around since that long, try running `npm info proj4` and see what the dist-tags show

theashyster:
@calvinmetcalf Okay, yeah, found the issue thanks to the `npm info proj4` command. I had forgotten that we used a local `npm` repository and that is not yet refreshed with the latest versions available. Thanks for the help, once again. 👍

calvinmetcalf:
ok this is all published as 2.4.0

PR #225, opened 2017-01-24
fix adjust_zone

calvinmetcalf:
quick fix to the adjust_zone function it looks like most of the time the original version wouldn't actually return anything ping @theashyster

### DIFF
--- a/lib/common/adjust_zone.js
+++ b/lib/common/adjust_zone.js
@@ -1,17 +1,16 @@
 var adjust_lon = require('./adjust_lon');
 
 module.exports = function(zone, lon) {
-  if (!zone) {
+  if (zone === undefined) {
     zone = Math.floor((adjust_lon(lon) + Math.PI) * 30 / Math.PI);
 
     if (zone < 0) {
       return 0;
-    }
-    else if (zone >= 60) {
+    } else if (zone >= 60) {
       return 59;
     }
-  }
-  else {
+    return zone;
+  } else {
     if (zone > 0 && zone <= 60) {
       return zone - 1;
     }
--- a/test/testData.js
+++ b/test/testData.js
@@ -325,6 +325,11 @@ var testPoints = [
     ll: [2, 1],
     xy: [1057002.4052152266, 110955.14117382761]
   },
+  {
+    code: '+proj=utm +lon_0=-3 +ellps=GRS80 +lat_1=0.5 +lat_2=2 +n=0.5',
+    ll: [2, 1],
+    xy: [1057002.4052152266, 110955.14117382761]
+  },
   // these test cases are related to the original issue on GitHub
   {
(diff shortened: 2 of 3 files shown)


theashyster:
For me it looked like only the changes you made in `utm.js` where required for this to work, but the changes for `adjust_zone.js` also makes a lot of sense. So looks good to me. 👍

lib/common/adjust_zone.js
```diff
@@ -1,17 +1,16 @@
...
     if (zone < 0) {
       return 0;
-    }
-    else if (zone >= 60) {
+    } else if (zone >= 60) {
       return 59;
     }
-  }
-  else {
+    return zone;
```
calvinmetcalf: without this being here the only zones that can be added are 0 or 59

theashyster: Yes, that is correct, I somehow overlooked it. Great 👍

theashyster: I wrote about only the changes in `utm.js`, because when I changed the check in my `etmerc` branch it worked fine with the zone 1, but I see now that I did not look deep enough. So thank you for noticing this, because this could cause a separate kind of bug. Can you add a unit test for this?

test/testData.js
```diff
@@ -325,6 +325,11 @@ var testPoints = [
     ll: [2, 1],
     xy: [1057002.4052152266, 110955.14117382761]
   },
+  {
+    code: '+proj=utm +lon_0=-3 +ellps=GRS80 +lat_1=0.5 +lat_2=2 +n=0.5',
```
calvinmetcalf: this zone 30, but I have to put in the lon_0 that corresponds to zone 29, which seems wrong unless I'm missing something

lib/common/adjust_zone.js
calvinmetcalf: see comment bellow but it seems like we'd want to return zone-1 here

theashyster: The `mapshaper-proj` works with just `zone` here, and I took the code from there, basically, so it should work with `zone`.

test/testData.js
calvinmetcalf: hm never mind zone 30 is -3, it looks like its ok, will merge then

calvinmetcalf:
ok published as 2.3.17

test/testData.js
theashyster: Hmm, I just run some test code and it looks like this is how it is supposed to be. Using `+lon_0=-3` you set in what is supposed to be `zone=30`, but is calculated to be `zone=29`. The same happens with `+proj=utm +zone=33 +datum=WGS84 +units=m +no_defs`, we specify `zone=33`, but it will be calculated to be `zone=32`.

PR #226, opened 2017-01-24
Add etmerc projection to proj4js and use it for utm

theashyster:
Reason of doing this is because calculation results for `proj4js` and `cs2cs` 4.9.3 (that now uses `etmerc`) differs and it means that it can be improved using the `etmerc` projection. Suggestion by @mbloch on the original issue #187

Hugely inspired by `mapshaper-proj` and how `etmerc` projection is calculated there https://github.com/mbloch/mapshaper-proj/blob/master/src/projections/etmerc.js

Using `etmerc` gives `UTM` a really great precision and is a generally suggested change, `proj.4` starting from version 4.9.3 has also switched for `UTM` to use `etmerc`.

Here is the comparison sheet between `proj4js` with `etmerc`, `mapshaper-proj` using `etmerc` and `proj.4` using `etmerc` (c library) https://docs.google.com/spreadsheets/d/1W6tqCLL54q1rSfXKQOHmqQ-7jxnptbXyDFlrVX3mkH4/edit?usp=sharing

Ping @calvinmetcalf


lib/projections/etmerc.js
```diff
@@ -0,0 +1,161 @@
...
+  this.x0 = this.x0 !== undefined ? this.x0 : 0;
+  this.y0 = this.y0 !== undefined ? this.y0 : 0;
+  this.long0 = this.long0 !== undefined ? this.long0 : 0;
+  this.lat0 = this.lat0 !== undefined ? this.lat0 : 0;
+
+  var f = this.es / (1 + Math.sqrt(1 - this.es));
+  var n = f / (2 - f);
+  var np = n;
+
+  cgb[0] = n * (2 + n * (-2 / 3 + n * (-2 + n * (116 / 45 + n * (26 / 45 + n * (-2854 / 675 ))))));
```
calvinmetcalf: if these need to be calculated for each instance then they should be local to the object, otherwise you need a guard to make sure it's only done once and they are never modified

theashyster: Yeah, looks like they should be local to the object, since they might differ for each new proj4 instance with a different proj4 string if we create two separate instances, right? I will test that when I will get to my PC and fix it.

calvinmetcalf: yeah in with case they should jsut be this.cgb etc

theashyster: I agree, I have fixed it.

lib/projections/etmerc.js
```diff
@@ -0,0 +1,160 @@
...
+var hypot = require('../common/hypot');
+var asinhy = require('../common/asinhy');
+var gatg = require('../common/gatg');
+var clens = require('../common/clens');
+var clens_cmplx = require('../common/clens_cmplx');
+var adjust_lon = require('../common/adjust_lon');
+
+exports.init = function() {
+  if (this.es <= 0) {
+    return;
```
calvinmetcalf: probably just throw an error here

theashyster: Fixed it :)

theashyster:
Hmm, I don't have any `wkt` based `etmerc` examples on hand, but I can search the web for something.

theashyster:
@calvinmetcalf I added another `etmerc` test, but I cannot find any `wkt` based examples that uses `etmerc`.

calvinmetcalf:
so there are no epsg standardized projections that use etmerc ?

theashyster:
I actually don't know, I checked quite a lot of `UTM` projections here http://www.spatialreference.org/ref/epsg/25833/ but all of them had `Transverse_Mercator` as the projection in the wkt.

calvinmetcalf:
ok great then, I merged this into #224 so it will land with that one

theashyster:
@calvinmetcalf Ok, thanks a lot. Yeah, saw the discussion, great. Could it slow the possibility of a new release by much?

calvinmetcalf:
yes it will slow it a bit but shouldn't be bad

theashyster:
Ok, great to hear. Thanks a lot for the really quick feedback and help.

theashyster:
@calvinmetcalf How big is the possibility that a new release including these changes comes out till Monday / Tuesday?

calvinmetcalf:
If @ahocevar doesn't get back to me I can at least do a beta release, which might not be a terrible idea as this could theoretically break stuff.

theashyster:
@calvinmetcalf Okay, that sounds great. Yeah, that is true, and the beta release would also help to test this better.

ahocevar:
@calvinmetcalf you have my full +1 to do a release. I'd consider this an important fix, not a breaking change.

theashyster:
This is fantastic, great, heh :)

PR #228, opened 2017-01-26
Fix running istanbul on Windows

bartvanandel:
Because npm is aware of package executables, such as `istanbul`, using the full path is not necessary, moreover it is not platform independent. The test script (`npm test`) failed on Windows because of this. This issue can be resolved by simply calling `istanbul` without the full path.

Note that the "mocha" argument supplied to `istanbul` still requires the full path because of the way it is being used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "doc": "docs"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
this is fixed in #224

Issue #231, opened 2017-01-31
switch the ellipse and datum stuff to do a default export


calvinmetcalf:
the `import * as foo from './bar'` *looks* like an object, but it isn't really one and that may break with more complaint module bundlers

PR #232, opened 2017-02-01
Export default for objects

calvinmetcalf:
import * as foo from 'bar' don't import a real object, it's all special and we want to avoid that #231


lib/constants/grids.js
```diff
@@ -1,23 +0,0 @@
-// Based on . CTABLE  structure :
```
calvinmetcalf: right this was something old we were totally not using anymore

calvinmetcalf:
going to go ahead and merge this since nobody objected

DenisCarriere:
@calvinmetcalf I don't believe you are using the export `default` correctly for many of these libraries.

`export default` should be used if you are trying to export a single module with a single value.

If you need to export a collection of functions then you should be using `named exports`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

Example snippet from [`datum.js`](https://github.com/proj4js/proj4js/blob/975b7748cddf83514017e78eabadae720f09223d/lib/constants/Datum.js).

**Misuse of `default` export**
```javascript
var exports = {};
export {exports as default};
exports.wgs84 = {
  towgs84: "0,0,0",
  ellipse: "WGS84",
  datumName: "WGS84"
};

exports.ch1903 = {
  towgs84: "674.374,15.056,405.346",
  ellipse: "bessel",
  datumName: "swiss"
};
```
**Using named exports**

```javascript
const wgs84 = {
  towgs84: "0,0,0",
  ellipse: "WGS84",
  datumName: "WGS84"
};

const ch1903 = {
  towgs84: "674.374,15.056,405.346",
  ellipse: "bessel",
  datumName: "swiss"
};
export {wgs84, ch1903};
```

@andy-ms Can you weigh on this?

DenisCarriere:
> import * as foo from 'bar' don't import a real object, it's all special 

This is mostly used in Typescript, since you are using Rollup your imports should be:

```javascript
import foo from 'bar'
```

andy-ms:
Well, I don't want to tell people how to write their libraries... I think @calvinmetcalf  may be concerned that in an ES6 environment, given `import * as x from "foo"`, `x` can't be a function or class. A lot of libraries have the style of exporting a single function, with other exports being properties of that function.
In my opinion that's not a good style and you should have a default export that *is* a function/class, and named exports in addition to that, i.e. `import myFunction, { namedExport1, namedExport2 } from "foo";`. But again, this is someone else's library and I don't really have a stake in it.

calvinmetcalf:
so bear in mind that this has been ported from commonjs so it has some idioms that are probably not great and we wouldn't have chosen if we were doing it from scratch but they are what they are.  

wherever possible I've tried to use named exports, but in datum, we aren't actually exporting a collection of things, we're exporting an object that we are treating as an object not a collection of names that's why we are using a default export.  That being said it should have no bearing on the public API.

as for your comment @andy-ms having both named and default exports are fine, as long as you never have to interact with commmonjs modules, in which case everything is suddenly really complicated

DenisCarriere:
> That being said it should have no bearing on the public API.

This causes breaking change to Typescript/Angular2 libraries.

My only main concern is the [`index.js`](https://github.com/proj4js/proj4js/blob/975b7748cddf83514017e78eabadae720f09223d/lib/index.js) since this does change the way you import `proj4` as a libary.

**Before**

```javascript
import * as proj4 from 'proj4'
```
**Now**

```javascript
import proj4 from 'proj4'
```
@calvinmetcalf I'd be happy to send a PR your way, I'm currently work with libraries using Rollup.

DenisCarriere:
>  but in datum, we aren't actually exporting a collection of things, we're exporting an object that we are treating as an object not a collection of names that's why we are using a default export. 

The example I listed above https://github.com/proj4js/proj4js/pull/232#issuecomment-293277762 has the exact same functionality & outputs as your current implementation.

calvinmetcalf:
```
exports {foo, bar}
```
and
```
var exports = {foo,bar};
export {exports as default}
```
don't do the same thing, the top one must be imported with
```
import * as bat from 'bar';
```
while the bottom one is imported as 
```
import bat from 'bar';
```
I don't know how it works in typescrypt, I just know that in es6 modules, thats how it works

additionally `bat` object in those 2 examples are NOT the same, in one it's a vanilla object, in the other it's a [Module Namespace Exotic Objects](http://www.ecma-international.org/ecma-262/6.0/#sec-module-namespace-exotic-objects), or in cases like rollup it's a frozen or otherwise special object.

DenisCarriere:
@calvinmetcalf Thanks your in-depth explanations 👍 

I've looked into this further https://github.com/DefinitelyTyped/DefinitelyTyped/issues/15663#issuecomment-293402296 and there seems to be no issues with `Proj4`, just seems odd that two users have noticed the same issue (related to the `default` import/export issue mentioned above).

**Import/Export** 

```javascript
exports {foo, bar}
```
Is a Typescript thing... (ops 🤦‍♂️ ).

The ES6 equivalent would  be:

```javascript
export default {foo, bar};
```

As a conclusion... disregard the issues I mentioned above, I think there's still room for improvement from the commonjs port (not a deal breaker).

@calvinmetcalf Thanks for your patience 👍

andy-ms:
@DenisCarriere Actually, `export { foo, bar };` is ES6 [syntax](https://github.com/babel/babylon/blob/master/ast/spec.md#exportnameddeclaration).

DenisCarriere:
@andy-ms 👍 That's true, my mistake.

PR #233, opened 2017-02-14
Fix from_greenwich property name

tchandelle:


calvinmetcalf:
woops this looks good, though a test that catches this would be great

tchandelle:
@calvinmetcalf Here is a test. I made sure it fails without 2f549d0

calvinmetcalf:
thanks!

PR #242, opened 2017-03-03
fix ellipse issue

### LINKED ISSUE #240: Unable to derive sphere constant when projection ellps is undefined
This was introduced in in [this commit](https://github.com/proj4js/proj4js/commit/9dc0fb7921fe697db2a8beb2fc7abf01f2cb1e5d) when the map lookup on `Ellipsoid` was changed to use the new [match function](https://github.com/proj4js/proj4js/blob/master/lib/match.js#L2)

[Previously the map lookup](https://github.com/proj4js/proj4js/blob/8c0c0eef0bfbf74ee82c2463e1adab65faea02a4/lib/deriveConstants.js#L32) would return `undefined` which would then cause `ellipse` to default to `Ellipsoid.WGS84`. Now if the key passed to `match` is `undefined`, it causes a TypeError to be thrown [during the call to key.toLowerCase()](https://github.com/proj4js/proj4js/blob/master/lib/match.js#L7).

I'm not sure if this new behavior is intended and is only now causing our own test cases to fail or if `match` should return `undefined` early if no key is passed.

calvinmetcalf:
basically #241 but I put the ellipse check in a slightly different spot where we do similar checks, fixes #240

### DIFF
--- a/lib/Proj.js
+++ b/lib/Proj.js
@@ -35,7 +35,7 @@ function Projection(srsCode,callback) {
   }
   json.k0 = json.k0 || 1.0;
   json.axis = json.axis || 'enu';
-
+  json.ellps = json.ellps || 'wgs84';
   var sphere_ = dc_sphere(json.a, json.b, json.rf, json.ellps, json.sphere);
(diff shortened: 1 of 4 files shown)


fredj:
Thanks @calvinmetcalf

Issue #237, opened 2017-02-27
UTM regression in 2.3.16/2.3.17 


MrSnyder:
I noticed a (tiny) numerical issue when updating from 2.3.15 to 2.3.17. My use case is converting from WGS 84 to EPSG:32631 (UTM zone 31):

Input: [3, 0] (center meridian of zone 31)
Actual output: UTM, Zone 31: [500000.000000001, 0]
Expected output: [500000, 0]

I know this is a minimal difference, but AFAIK, the center meridian of each UTM zone is defined to be exactly at 500 km, right?

I prepared two fiddles to demonstrate what I believe to be a regression between 2.3.15 and 2.3.17:

* 2.3.17: https://jsfiddle.net/d52j2nkm/1/
* 2.3.15: https://jsfiddle.net/d52j2nkm/2/

What's also strange is that the regression only seems to affect odd UTM zones, for even zones, the meridian is still exactly at 500km:

* 2.3.17, odd zone: https://jsfiddle.net/d52j2nkm/3/

calvinmetcalf:
2 things

1. thats just a floating point rounding error
2. check against alpha by doing npm install proj4@alpha it's probably already fixed

MrSnyder:
calvinmetcalf, thanks for your super-quick reply.

1. Yes, I know. And I don't want to annoy anybody here, still cs2cs on the command line gives exactly 500km (as does 2.3.15). Plus it works for even UTM zones. So I figured it would be worth pointing this out.
2. Thanks for the hint. I did it just now, but it's still there.

calvinmetcalf:
ok so this is almost certainly related to the UTM we put in around then so I'll take a look

calvinmetcalf:
ok checkout #243

MrSnyder:
Very nice. I checked with current master, and I get the exact value of [500000, 0] again. Fixed by commit 7d94169bb0fa1ab1315176446348979367a575a1.

Thank you.

PR #243, opened 2017-03-03
fix floating point reversion (#237)

calvinmetcalf:
paging @theashyster as this effects some code he wrote, basically what he did involved a formula that for some reason required the zone to be one lower then it really was but also seemed to have a floating point error in it, so this reverts to the old formula which doesn't require the zone to be one less so that's why it touches the the adjust lon stuff, I probably need some tests to check zones 0 and 60 while I'm at it

### DIFF
--- a/lib/common/adjust_zone.js
+++ b/lib/common/adjust_zone.js
@@ -2,17 +2,13 @@ import adjust_lon from './adjust_lon';
 
 export default function(zone, lon) {
   if (zone === undefined) {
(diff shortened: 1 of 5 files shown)


theashyster:
@calvinmetcalf I tested the 2.4.3 release a little bit and it seems to work fine for me. Okay, what I did was pretty much take the code written by @mbloch and adapt it for `proj4js` library, so I might not know the complete reasoning behind why it was done that way. Thanks for catching the error and fixing it, I hope it won't bring out some unexpected problems in the future. :)

PR #245, opened 2017-03-16
bring over all properties (fixes #230)

### LINKED ISSUE #230: Projection from/to the same CRS gives large Z difference since 2.3.16
Example using EPSG:7415 (Amersfoort RD New + NAP), tested in PhantomJS

Head:
``` .js
const proj4 = require('./dist/proj4.js');
proj4.defs('RD', '+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.999908 +x_0=155000 +y_0=463000 +ellps=bessel +units=m +towgs84=565.2369,50.0087,465.658,-0.406857330322398,0.350732676542563,-1.8703473836068,4.0812 +no_defs');
```

Using [2.3.16](releases/2.3.16) or [2.3.17](releases/2.3.17):
``` .js
proj4('RD', 'RD', [100000, 437000, 10]);
// {
//    "0": 100000.00006519399,
//    "1": 437000.00057943614,
//    "2": -43.74852570705116
// }
```

Must be a regression, because [2.3.15](releases/2.3.15) gives these results:
``` .js
proj4('RD', 'RD', [100000, 437000, 10]);
// {
//    "0": 100000.00009487881,
//    "1": 436999.9997790383,
//    "2": 10.000132272019982
// }
```

Now, ideally `proj4(a, a, b)` should probably recognize that the "from" and "to" CRS are the same and just return `b`, but even without this check, I don't think such large changes in Z are acceptable. What's going on here?

calvinmetcalf:


calvinmetcalf:
92f412f

PR #246, opened 2017-03-24
An included Typescript definitions file

pierremarc:
As stated in Typescript documentation [1], it's favourable to
have the definitions file alongside the source.

As far I know, definition files on _Definitely Typed_  aren't versioned and it's easy to end up using a version of an NPM package that doesn't match its definitions on DT -- I've been there :)

The file here is based on the work done by @DenisCarriere and I'd be glad to hear your opinion on such a setup. 

[1] http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Proj4js is a JavaScript library to transform point coordinates from one coordinate system to another, including datum transformations.",
   "main": "dist/proj4-src.js",
   "module": "lib/index.js",
+  "typings": "proj4.d.ts",
   "directories": {
     "test": "test",
     "doc": "docs"
--- a/proj4.d.ts
+++ b/proj4.d.ts
@@ -1,0 +1,67 @@
+// Type definitions for proj4 2.3
+// Project: https://github.com/proj4js/proj4js
+// Definitions by: Denis Carriere <https://github.com/DenisCarriere>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare namespace proj4 {
+    type TemplateCoordinates = number[] | InterfaceCoordinates;
+
+    interface InterfaceCoordinates {
+        x: number;
+        y: number;
+        z?: number;
+        m?: number;
+    }
+
+    interface InterfaceDatum {
+        datum_type: number;
+        a: number;
+        b: number;
+        es: number;
+        ep2: number;
+    }
+
+    interface Static {
+        forward(coordinates: InterfaceCoordinates): InterfaceCoordinates;
+        inverse(coordinates: InterfaceCoordinates): InterfaceCoordinates;
+    }
+
+    interface InterfaceProjection {
+        datum: string;
+        b: number;
+        rf: number;
+        sphere: number;
+        es: number;
+        e: number;
+        ep2: number;
+        forward(coordinates: InterfaceCoordinates): InterfaceCoordinates;
+        inverse(coordinates: InterfaceCoordinates): InterfaceCoordinates;
+    }
+
+    export const defaultDatum: string;
+
+    export function Proj(srsCode: any, callback?: any): InterfaceProjection;
+
+    export const WGS84: any;
+
+    /**
+     * Depecrated v3
+     */
+    export function Point(x: number, y: number, z?: number): InterfaceCoordinates;
+    export function Point(coordinates: TemplateCoordinates | string): InterfaceCoordinates;
+
+    export function toPoint(array: number[]): InterfaceCoordinates;
+
+    export function defs(name: string, projection?: string): any;
+    export function defs(name: string[][]): any;
+
+    export function transform(source: InterfaceProjection, dest: InterfaceProjection, point: TemplateCoordinates): any;
+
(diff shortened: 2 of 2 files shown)


calvinmetcalf:
so none of the maintainers, as far as I know, are really experts in this and able to maintain it is the main issue i have

calvinmetcalf:
also note the caveat on the typescript documentation 

> If you control the npm package you are publishing declarations for, then the first approach is favored.

pierremarc:
I understand. OTOH, the file is kind of simple and should be easy to maintain. 
Actually, I'm no expert either but could dedicate some time to maintain it as we use it for a long term project and I'd rather prefer to maintain this than a fork.

calvinmetcalf:
I really think that it being separate sends appropriate signal about the level of support by the proj4 maintainers. 

Is this an actual problem or a theoretical problem based on how things *should* be?

pierremarc:
Its

pierremarc:
Well, I've got to check again if its me doing wrong, but it seems types/proj4 is not in sync, accepting number[] where only {x:number, y: number} should be. For me it led to non-crashing unprojected coordinates.
But again I'll need to check, can't do it now really from my mobile phone walking.

calvinmetcalf:
no it can take either an array or an object

DenisCarriere:
I would recommend you look at TurfJS and look at how the Typescript definitions were done there.
https://github.com/Turfjs/turf/tree/master/packages/turf-helpers

I'm very 👍 having the `index.d.ts` within the source code, however you need to have `proj4js` continue to maintain the the Typescript definitions.

If no one is willing to maintain it, then 👎 and you should let DefinitelyTyped maintain it.

DenisCarriere:
@pierremarc You can always send a [PR to DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/proj4) with any fixes to [`@types/proj4`](https://www.npmjs.com/package/@types/proj4), the Microsoft folks provide very good feedback when you send a PR.

ahocevar:
I'm also in favor of providing the TypeScript definitions at DefinitelyTyped.

pierremarc:
I understand that maintaining a definitions file within the source code is yet another piece to take care of, not fun.
On my side, the problem was from (mis)using forward/inverse on Proj instances as in 

```javascript
import * as proj4 from 'proj4';
const p: proj4.InterfaceProjection = proj4.Proj('EPSG:3857');
const result = p.forward([0, 0]);
console.log(result);
// >> [ 0, 0, x: NaN, y: NaN ]
```

And yes, I understand that's not supposed to be used like this -- porting from es5 is a long path :) nonetheless it compiles fine and that's a problem. 

If there's no interest in maintaining the definitions as part of proj4 itself, well I guess it will remain on DefinitelyTyped, which is an OK solution. 

thanks for the feedback

@DenisCarriere I can update your definitions, but TBH if you've got everything setup it might be easier that I send a patch to you?

DenisCarriere:
@pierremarc no, don't send it to me, the whole reason why DT exists is to host communal TypeScript definitions. Just send a PR to them.

DT will send me an email with the proposed changes and I just need to 👍 / 👎  the PR.

wulfsberg:
The unversioned nature of DefinitelyTyped is still an issue, though, which is one major point in favor of bundling it with the (versioned) project, so the definitions are actually in sync with the code.

calvinmetcalf:
@wulfsberg the api hasn't changed, the previous version was just wrong

calvinmetcalf:
closing this as the consensus is that none of the authors of proj4 particularly want to/know how to maintain a type definition so it should live in definitely typed

wulfsberg:
@calvinmetcalf I just want to point out that the API *did* change, with the ES6 default exports. 2.3 and 2.4 are not compatible when it comes to imports, and it will not be possible to provide a DefinitelyTyped version which supports both versions.

But the choice to rely on DefinitelyTyped is perfectly valid, so I am ok with that. DefinitelyTyped is exactly for this type of situations, where the maintainers of a non-TypeScript project shouldn't be bothered with authoring definition files just because some of us like to use it that way.

@pierremarc Did you get any further with a pull request to DefinitelyTyped? If the API is otherwise unchanged, it is, AFAICS, purely a matter of changing the `export = proj4;` to `export default proj4;` (and import it with `import proj4 from 'proj4';` in the test). I can do that if you want.

pierremarc:
@wulfsberg 
To be honest, I'm such in a rush at the moment that I didn't take the time to prepare a proper PR for DT and just use my fork for my projects. And yes  ```export default``` did the trick :)

wulfsberg:
Following up on this, from comments on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16637:
While the release includes a rollup bundle with ES6 default exports (which it seems a webpack environment eagerly picks up, hence the effective change in signature), the actual main entry of the package is still an UMD module.
So the trick is to use the ES6-style `import proj4 from 'proj4';` and let your TypeScript compiler `"allowSyntheticDefaultImports": true` to pick up the UMD export.

PR #251, opened 2017-04-25
Check sanity of inputs

### LINKED ISSUE #250: Inconsistency when transforming NaN, Infinity using various projections
There seems to be an issue when the input point is something like `[NaN, NaN]` or `[Infinity, Infinity]` and we try to call transform using some of the projections. The returned result is quite inconsistent - some times it returns `[NaN, NaN]`, some times `[Infinity, Infinity]`, some times coordinates like `[0, 90]` or `[45, 90]`.

When we pass in such params as `NaN` or `Infinity` shouldn't there be some safe guard checks in front of the actual transformation that checks if the passed in points are valid to work with? At least check if they are not `NaN` or `Infinity` since using transformation on such values won't give any meaningful results?

I have created 2 separate JS Bin examples
https://jsbin.com/fipuxiqufe/edit?html,console,output (using proj4js 2.4.3)
https://jsbin.com/mapotalape/edit?html,console,output (using proj4js 2.3.15)

I created 2 separate examples, because I also wanted to show the issue with `tmerc` projection, that I have introduced. When using 2.3.15 and passing in `[NaN, NaN]` to the `inverse` function the returned result is also `[NaN, NaN]`. But since 2.3.16 when passing `[NaN, NaN]` to the `inverse` function the returned result is `[0, 90]` and I don't think this is correct behavior.

Also after the introduction of `etmerc` in 2.3.18-alpha, when passing in `[NaN, NaN]` to the `inverse` function of `utm` projection (which uses `etmerc`) the returned result is `[Infinity, Infinity]`, but I am not sure if that should be the case or not.

I already looked at the code of `tmerc` and `etmerc` and we set those values there when some conditions are not met for the result, but I don't think those conditions were put there to also catch such things as `NaN` or `Infinity`.

So all in all would be great to hear some advice and thoughts from you, guys.

calvinmetcalf:
fixes #250


lib/check-sanity.js
```diff
@@ -0,0 +1,21 @@
+export default function (point) {
+  checkCoord(point.x);
+  checkCoord(point.y);
+}
+function checkCoord(num) {
+  if (typeof Number.isFininte === 'function') {
```
theashyster: Typo, should be `isFinite`

lib/check-sanity.js
```diff
@@ -0,0 +1,21 @@
...
+    }
+    throw new TypeError('coordinates must be finite numbers');
+  }
+  if (typeof num !== 'number') {
+    throw new TypeError('coordinates must be finite numbers');
+  }
+  if (num !== num) {
+    throw new TypeError('coordinates must be finite numbers');
+  }
+  if (!isFinite(num)) {
```
theashyster: Don't you already check this at the top?

lib/check-sanity.js
```diff
@@ -0,0 +1,21 @@
...
+  if (typeof Number.isFininte === 'function') {
+    if (Number.isFininte(num)) {
+      return;
+    }
+    throw new TypeError('coordinates must be finite numbers');
+  }
+  if (typeof num !== 'number') {
+    throw new TypeError('coordinates must be finite numbers');
+  }
+  if (num !== num) {
```
theashyster: Would this ever happen?

lib/check-sanity.js
```diff
@@ -0,0 +1,21 @@
+export default function (point) {
+  checkCoord(point.x);
+  checkCoord(point.y);
+}
+function checkCoord(num) {
+  if (typeof Number.isFininte === 'function') {
+    if (Number.isFininte(num)) {
+      return;
+    }
+    throw new TypeError('coordinates must be finite numbers');
```
theashyster: This error could be saved to a constant to avoid duplication

lib/check-sanity.js
calvinmetcalf: good chatch

lib/check-sanity.js
calvinmetcalf: this is a check for NaN, basicaly we try to use the new Number.isFinite method which checks that the number isn't nan, that it's not Infinity, and that it's actually a number, if that isn't available we do them seperatly

lib/transform.js
```diff
@@ -3,6 +3,8 @@ import datum_transform from './datum_transform';
 import adjust_axis from './adjust_axis';
 import proj from './Proj';
 import toPoint from './common/toPoint';
+import checkSanity from './check-sanity';
```
theashyster: No other files in this library is named using `-` in them

lib/check-sanity.js
calvinmetcalf: Number.isFinite checks that it's a number, it's not NaN and that it's not Infinity or -Infinity, the global isFinite function is older, coerces to a number and returns true for NaN.

test/test.js
```diff
@@ -204,6 +204,26 @@ function startTests(chai, proj4, testPoints) {
           new proj4.Proj('fake one');
         }, 'fake one', 'should work');
       });
+      it('should throw when passed a null', function() {
+        assert.throws(function() {
+            proj4('+proj=utm +zone=31', [null, 0]);
+        }, 'coordinates must be finite numbers', 'should work');
+      });
+      it('should throw when passed NAN', function() {
```
theashyster: NAN -> NaN

test/test.js
```diff
@@ -204,6 +204,26 @@ function startTests(chai, proj4, testPoints) {
           new proj4.Proj('fake one');
         }, 'fake one', 'should work');
       });
+      it('should throw when passed a null', function() {
```
theashyster: a null -> null

lib/check-sanity.js
calvinmetcalf: not if we want the stack traces to point correctly, also errors are mutable

lib/check-sanity.js
theashyster: Okay, you just use the older on if the newer one does not exist to be sure, cool.

lib/check-sanity.js
theashyster: Okay, yep, that makes sense.

lib/check-sanity.js
theashyster: I mean, the error message, not the error itself. :)

lib/check-sanity.js
calvinmetcalf: well if the newer one doesn't exist, we have to do 3 checks to get the same functionality including this one

theashyster: Yep, I eventually came to that conclusion, thanks.

theashyster:
Looks good to me 👍

lib/check-sanity.js
```diff
@@ -9,13 +9,7 @@ function checkCoord(num) {
...
     throw new TypeError('coordinates must be finite numbers');
   }
-  if (typeof num !== 'number') {
-    throw new TypeError('coordinates must be finite numbers');
-  }
-  if (num !== num) {
-    throw new TypeError('coordinates must be finite numbers');
-  }
-  if (!isFinite(num)) {
+  if (typeof num !== 'number' || num !== num || !isFinite(num)) {
```
michie1: What's the purpose of `num !== num`? Isn't this always returning false?

calvinmetcalf:
it's true for NaN

PR #257, opened 2017-06-13
fix mollweide projection infinite loop

### LINKED ISSUE #256: Infinite operation on Mollweide projection
Apologize in advance if this is hard to replicate.

Working version: 2.3.17
Bugged version: 2.4.x

I had a bug where drawing a d3 vector of the world projected onto a Leaflet tile map was hanging with Mollweide projection. Specifically the `forward` function. Tracked it down to this [code](https://github.com/proj4js/proj4js/blob/master/lib/projections/moll.js#L19):

```
/* Iterate using the Newton-Raphson method to find theta
      -----------------------------------------------------*/
  for (var i = 0; true; i++) {
    var delta_theta = -(theta + Math.sin(theta) - con) / (1 + Math.cos(theta));
    theta += delta_theta;
    if (Math.abs(delta_theta) < EPSLN) {
      break;
    }
  }
```

The problem was that `Math.abs(delta_theta)` was never less than `EPSLN`.
I solved it by changing the for loop to end after 100 iterations `for (var i = 0; i<100; i++) {` which is obviously a hack, but seems to work well enough for my purposes. 

Logs leading up to the hang:
```
console.log(lon, lat, delta_lon, theta, con);
-3.141592653589793 -1.3981832637726574 -3.141592653589793 -1.3981832637726574 -3.094906445571509
```

```
console.log(EPSLN);
2.220446049250313e-16
```

```
if (i < 100) console.log(delta_theta);
-0.6072792942381932
-0.315141935607568
-0.1333818531884728
-0.027420655741693258
-0.0011228675542062185
-0.0000018450944152329483
-4.976677737435555e-12
-2.120442154881964e-15
2.120442154881978e-15
-2.120442154881964e-15
2.120442154881978e-15
```

I don't know the math well enough to figure out why exactly this starts alternating values, but because it does, it never drops below EPSLN.

calvinmetcalf:
fixes #256

### DIFF
--- a/lib/constants/values.js
+++ b/lib/constants/values.js
@@ -10,7 +10,10 @@ export var SIXTH = 0.1666666666666666667;
 export var RA4 = 0.04722222222222222222;
 /* 17/360 */
 export var RA6 = 0.02215608465608465608;
-export var EPSLN = (typeof Number.EPSILON === 'undefined') ? 1.0e-10 : Number.EPSILON;
+export var EPSLN = 1.0e-10;
+// you'd think you could use Number.EPSILON above but that makes
+// Mollweide get into an infinate loop.
+
 export var D2R = 0.01745329251994329577;
 export var R2D = 57.29577951308232088;
 export var FORTPI = Math.PI/4;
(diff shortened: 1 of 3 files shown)


bartvde:
Is this planning to be merged?

bartvde:
thanks @calvinmetcalf

PR #258, opened 2017-06-16
Fix error when geodeticToGeocentric returns null

flexf1t:
JS execution stops when geodeticToGeocentric returns null point and throws exception that p is null. 
I think it will be much better if it returns somehow like in c++ proj4 function pj_geodetic_to_geocentric.c:409.
At least in openlayers when View proj is EPSG:28405 but Layer proj is EPSG:3857 it will draw something, not just stop on exception.
Without fix:
ol 28405 bug proj
With fix:
ol 28405


Hanmac:
happens for me with: "EPSG:31467 (GK 3)"
i would like to see this fix merged

msolvaag:
Please merge this!

PR #267, opened 2017-10-17
fix float precision issue

ksokolowski7:
There is an issue that long0 is NaN for some cases when result of gl * Math.tan(this.gamma0) is near 1. In JS is it possible that the result is 1.00...002 when correct value is 0.99..998. Math.asin can accept only values from -1 to 1.

I am not sure if we should take into account the value -1. 

Tested projection:

PROJCS["CH1903+_LV95",GEOGCS["GCS_CH1903+",DATUM["D_CH1903+",SPHEROID["Bessel_1841",6377397.155,299.1528128],TOWGS84[674.374,15.056,405.346,0,0,0,0]],PRIMEM["Greenwich",0],UNIT["Degree",0.0174532925199433]],PROJECTION["Hotine_Oblique_Mercator_Azimuth_Center"],PARAMETER["False_Easting",2600000],PARAMETER["False_Northing",1200000],PARAMETER["Scale_Factor",1],PARAMETER["Longitude_Of_Center",7.43958333333333],PARAMETER["Latitude_Of_Center",46.9524055555556],PARAMETER["Azimuth",90],UNIT["Meter",1,AUTHORITY["EPSG","9001"]],AUTHORITY["EPSG","2056"]]


calvinmetcalf:
do you mind adding that as a test? just add it [here](https://github.com/proj4js/proj4js/blob/029fa8cf6bd4d166c37007c143ae2599f566ec49/test/testData.js#L518)

ksokolowski7:
I added test data to testData.js. Of course it passes witch changes in [wkt-parser](https://github.com/proj4js/wkt-parser/pull/4) 
Could you merge my changes in wkt-parser?

calvinmetcalf:
I published it, and rebased your pull so hopfully the tests would pass but it's not showing up yet

ksokolowski7:
Can we rerun tests? Looks that previous build was cancelled on CI without any result.

PR #275, opened 2017-11-25
Add robinson projection

havlicekj:
I have added robinson projection (with tests).
Note that the results are slightly different than what cs2cs would produce, the reason being that the polynomial coefficients in https://github.com/OSGeo/proj.4/blob/master/src/PJ_robin.c#L20 are stored as single-precision floats, but JavaScript uses double precision. When I recompiled proj4 code to use doubles, the results matched exactly.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,7 +26,8 @@ var projs = [
   'vandg',
   'aeqd',
   'ortho',
-  'qsc'
+  'qsc',
+  'robin'
 ];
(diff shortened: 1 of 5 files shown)


calvinmetcalf:
I like this, @ahocevar your opinion ?

havlicekj:
Hi, what can I do to get this merged?

ahocevar:
I like it too @calvinmetcalf.

Issue #193, opened 2016-06-09
Issue with longlat -> stereo translations on non-earth spheroid


malaretv:
See here for an example:
http://mistakes.io/#398484798d88de68c7f9793edf5fd407

Translating between longlat and polar stereographic on the mercury spheroid doesn't seem to return consistent results.

calvinmetcalf:
when a this library doesn't really have create support for custom spheroids, also fyi non-earth spheroids of this solar system have flipped x axises which I doubt this library can handle.

malaretv:
Interesting.. have not had issues with custom spheroids translating between eqc, sinusoidal and longlat which is why I brought the issue up. I'll take this as an invitation to learn a bit about the math involved and investigate why the translations work in proj4 but not proj4js. :)

Never encountered the issue with flipped x axises but i'll keep in an eye out for it thanks.

swalterfub:
Problem still exists... any further investigations? I'm converting Mars stereographic to longlat coordinates...

swalterfub:
Looks like we have a bug in the implementation of the stereographic projection in proj4js... the calculation of c has a typo in line 3966 of the current proj4js-src.js (version 2.2.1).
Instead of:
`var c = 2 * Math.atan(rh / (0.5 * this.a * this.k0));` it should be:
`var c = 2 * Math.atan(rh / (2 * this.a * this.k0));`
acoording to Snyder 1987: Map Projections - a Working Manual (https://pubs.usgs.gov/pp/1395/report.pdf) , formula 21-15.
Some testing of random values back and forth confirm the correction.

PR #281, opened 2017-12-22
Fixed calculation for angular distance c, see Snyder 1987, formula 21-15

swalterfub:
Described in Issue #193. For the correct formula, see 21-15 on page 159 of Synder: https://pubs.usgs.gov/pp/1395/report.pdf

### DIFF
--- a/lib/projections/stere.js
+++ b/lib/projections/stere.js
@@ -103,7 +103,7 @@ export function inverse(p) {
   var lon, lat, ts, ce, Chi;
   var rh = Math.sqrt(p.x * p.x + p.y * p.y);
   if (this.sphere) {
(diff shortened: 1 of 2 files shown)


calvinmetcalf:
a test would be great so we could avoid regressing this, otherwise this looks great

swalterfub:
Test is added. Is it what you meant? The coordinates in the test are retrieved by the proj command line utility

calvinmetcalf:
yes, will take a look

malaretv:
Indeed this fixes the issues i was having in #193. Thanks @swalterfub !!

Is there anything keeping this from being merged in I could help out with?

PR #291, opened 2018-07-17
added support for Israel 1993 / Israeli TM Grid

ksokolowski7:
Added support for Israel 1993 / Israeli TM Grid (http://epsg.io/2039).

Should be added together with https://github.com/proj4js/wkt-parser/pull/8


calvinmetcalf:
tests ?

ksokolowski7:
added

ksokolowski7:
I discovered that this projection still is shifted a bit. 

How _towgs84_ should be set? I did some tests from different tools and results are different.
For example for data x: 184808.119074685, y: 675054.042663833 the results are as below:

Tool | X | Y
------------ | -------------| -------------
epsg.io|34.8371672|32.16849
mygeodata.cloud/cs2cs/|34.83716723 | 32.16849013
tatuk calculator|34.83708894|32.16843036
arg gis api|34.8370958654819|32.1684216470369

The difference is around 10 meters. When I set _towgs84_ on "-35,56,52" position looks good compared to tatuk component. However I set this value by trial and error method.

PR #300, opened 2018-09-28
Fixes #279: fixed interpretation of axis order

### LINKED ISSUE #279: Axis order is not correctly interpreted
In adjust_axis, only the sign of each axis is taken into account, not their eventual ordering (e.g. enu is considered the same as neu)

### LINKED ISSUE #343: proj4 doesn't transform axis. axis ignore
i use proj4.defs(epsg, wkt)，wkt = "....AXIS["Geodetic latitude", NORTH], AXIS["Geodetic longitude", EAST];" . but proj4 dosen't transform AXIS to axis.  the openlayer need to use axis. the value of axis, such as "neu" or "enu". so if proj4 doesn't transform. openlayer not be able to use the AXIS. 

image

mbarto:
Depends on https://github.com/proj4js/wkt-parser/pull/10

Fixes #343.

### DIFF
--- a/lib/adjust_axis.js
+++ b/lib/adjust_axis.js
@@ -10,11 +10,20 @@ export default function(crs, denorm, point) {
     }
     if (i === 0) {
       v = xin;
+      if ("ew".indexOf(crs.axis[i]) !== -1) {
       t = 'x';
+      } else {
+          t = 'y';
+      }
+
     }
     else if (i === 1) {
       v = yin;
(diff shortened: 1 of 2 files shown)


calvinmetcalf:
a test would be helpful, it will obviously fail but I can test it against the other repo and publish them both

ahocevar:
See #344 for an updated version of this pull request.

PR #316, opened 2019-03-26
Addition to #230, added support for array coords

gpcboekema:
The fix in #245 as fix for #230 is not working for all cases.
This fix also adds a solution when using an array as coords (with z or z and m value).

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -6,15 +6,7 @@ function transformer(from, to, coords) {
   var transformedArray, out, keys;
   if (Array.isArray(coords)) {
     transformedArray = transform(from, to, coords);
-    if (coords.length === 3) {
-      return [transformedArray.x, transformedArray.y, coords[2]];
-    }
-    else if (coords.length === 4) {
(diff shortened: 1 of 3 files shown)


calvinmetcalf:
we probably want to generalize this for any array longer then 2

lib/core.js
```diff
@@ -6,12 +6,7 @@ function transformer(from, to, coords) {
   var transformedArray, out, keys;
   if (Array.isArray(coords)) {
     transformedArray = transform(from, to, coords);
-    if (coords.length === 3) {
-      return [transformedArray.x, transformedArray.y, transformedArray.z];
-    }
-    else {
-      return [transformedArray.x, transformedArray.y];
-    }
+    return [transformedArray.x, transformedArray.y].concat(coords.splice(2));
```
calvinmetcalf: why don't you put this where `return [transformedArray.x, transformedArray.y, transformedArray.z];` with the conditional being `coords.length > 2` that way the most common case isn't slowed down

Issue #313, opened 2019-01-29
Local projection result different from cs2cs


winsento:
**Proj4js** [jsfiddle](https://jsfiddle.net/winsent/v2k6onb5/)
```
var source = 'EPSG:4326';
var dest = '+proj=omerc +lat_0=51.25337665246619 +lonc=39  +k=0.9999406491557739 +x_0=96062.11582090713 +y_0=7596.58318533453 +gamma=-0.4017832147399834 +ellps=krass +towgs84=23.92,-141.27,-80.9,0,0.35,0.82,-0.12 +units=m +no_defs';
proj4(source, dest, [37.93, 51.72])
```
**cs2cs** from here https://mygeodata.cloud/cs2cs/ or QGIS

Results:
 - proj4js: `22238.8178835  60052.709825`
 - cs2cs:   `21872.8812039  59533.872365`

Issue #308, opened 2018-12-19
custom projection provides incorrect projection for omerc when compared with proj4


nethdeco:
Im have created a custom projection for a map created using point cloud data. The map is quite small around 500m x 500m.  The map needs to projected on top a google map. Therefore I created the following projection. 

+proj=omerc +lat_0=37.4769061 +lonc=141.0039618  +alpha=202.22  +k=1 +x_0=138 +y_0=77.65 +ellps=WGS84 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs

With the above projection the point cloud map could be overlayed on a google map through QGIS. 

However the pro4 gives the following result 
North 37.476802 East 141.003611
64.1736 168.2438

When tried with proj4js the results becomes the following 
North 37.476802 East 141.003611
4242217.031704698, 10384524.21017775

By analysing the parameters I found that both proj4 and proj4js gave same conversion values when the alpha value was between -90 and +90.  When the alpha value increases beyond that point the converted values are incorrect. 

I’m wondering if there is a work around to project using a alpha value beyond those values (above 90 and below 270) .

daniel-childs:
Hey @nethdeco, I've run into an issue with incorrect omerc alpha rotations as well. Lucky us! I'll see if I can get a patch made up in the next few days.

calvinmetcalf:
thanks @daniel-childs that would be welcome, though I don't know how quickly I'll be able to get to it over the holidays, i will try though

daniel-childs:
I had a cursory look through the code and it seems like it deviates from the original proj4 source by quite a bit. I'm not really up-to-date on the infrastructure in place so it's pretty difficult for me to track down where the error might be.
As a workaround I'm just using a server running "proj-bin" to do the conversions and piping it back to the browser.

winsento:
`+gamma` not implemented in this library.  You can see this in the implementation of [omerc](
https://github.com/proj4js/proj4js/blob/master/lib/projections/omerc.js).

**Oblique Mercator description**
https://github.com/OSGeo/proj.4/blob/master/docs/source/operations/projections/omerc.rst

**Omerc in proj4**
https://github.com/OSGeo/proj.4/blob/master/src/projections/omerc.cpp

Similar tickets #313 #301 #273 #308 #173 #146

Issue #301, opened 2018-10-10
+gamma is not correctly applied when different from +alpha in oblique Mercator projections


thedmi:
Consider the following two projections:

```
var p1 = "+proj=omerc +alpha=47.37757 +gamma=47.37757 +lonc=8.54175 +lat_0=47.37757 +lon_0=8.54175 +k_0=1 +towgs84=0,0,0 +units=m +x_0=0 +y_0=0 +ellps=WGS84 +no_defs";
var p2 = "+proj=omerc +alpha=47.37757 +gamma=137.37757 +lonc=8.54175 +lat_0=47.37757 +lon_0=8.54175 +k_0=1 +towgs84=0,0,0 +units=m +x_0=0 +y_0=0 +ellps=WGS84 +no_defs";
```

Both are oblique Mercator projections that only differ in the `gamma` parameter, which should result in a rotation of the XY-plane (by 90° in the example above). This leads us to the following transformations:

```
console.log(proj4(p1, [8.55499177954, 47.3775692352])); // Prints ~ [1000, 0] (ok)
console.log(proj4(p2, [8.55499177954, 47.3775692352])); // Prints ~ [1000, 0] (not ok, should be ~ [0, -1000])
```

So apparently the `gamma` parameter is not applied correctly.

The expected behavior can be verified e.g. with https://mygeodata.cloud/cs2cs/, or with proj4 directly.

thedmi:
This appears to be working correctly in current versions, I'll close the issue. Thanks a lot!

Issue #273, opened 2017-11-14
Wrong projection between WGS84 and GDM2000/Peninsular RSO


serge-markin:
The issue #173 which was closed is still there.

My code:
`var epsg4326 = '+proj=longlat +datum=WGS84 +no_defs'

var epsg3375 = '+proj=omerc +lat_0=4 +lonc=102.25 +alpha=323.0257964666666 +k=0.99984 +x_0=804671 +y_0=0 +no_uoff +gamma=323.1301023611111 +ellps=GRS80 +units=m +no_defs'

var transformation = proj4(epsg4326, epsg3375)
var input = [101.70979078430528, 3.06268465621428];
var output = transformation.forward(input);
// 744626.6428051005, -103618.62269607143
// expected: 412597.532715, 338944.957259`

I tried previous versions and got the same wrong results.

ceylanb:
+1

shanmugamg:
+1 
faced with similar problem and yet to find a solution

calvinmetcalf:
ok my first thought is that we aren't doing anything with the field gamma/rectified_grid_angle, but esri wkt files don't include that one so that's probably not it, looking at our tests we do have one for another omerc, but it's azimuth is only 90, I tested alaska state plane zone 1 and found that also doesn't work right, I suspect there has always been a bug in omerc, will look into it more

Ariki:
As  far as I understand, Proj4js seems to ignore `+gamma` parameter, whereas proj C library allows to specify it (by default it's assumed to be equal to `+alpha`). So projections that have gamma != alpha do not work correctly.

PR #317, opened 2019-04-15
Increase accuracy with omerc projection

dulldrums:
### NOTE:
_This implementation is based off of the mapshaper-proj implementation of [omerc](https://github.com/mbloch/mapshaper-proj/blob/master/src/projections/omerc.js), which is based directly off of the proj4 implementation of [omerc](https://github.com/OSGeo/proj.4/blob/master/src/projections/omerc.cpp). I attempted to match the structure of proj4js as best as possible._

Also, **one test I added is failing**. The test data I have added is coming from proj4, so it should be correct. It could indicate that there is more work to be done for this implementation.

This issue is mentioned in #308 , #313 , #301 , #273 , #308 , #173, and #146.

#### Problem:
The omerc projection is returning inaccurate values. As mentioned in #308, it appears as though the implementation is ignoring the `+gamma` value. Additionally, whenever the alpha values are +/-90, the transformed coordinates were inaccurate.

#### Solution:
I had found this implementation of [omerc](https://github.com/mbloch/mapshaper-proj/blob/master/src/projections/omerc.js) that appeared to be producing more accurate results than the proj4js implementation of omerc. There are several things in the mapshaper-proj implementation that are dependant on that library's infrastructure, so I tried to work around that as best as possible.

I would love to get a second set of eyes on this code to verify that this fix is actually helping. 

There is currently a failing test that I have added. 12 tests are failing with this implementation, 32 are failing with the previous one.

I'll continue to investigate why this test is failing, and update this PR as soon as I figure it out.

Thanks!

### DIFF
--- a/lib/projections/omerc.js
+++ b/lib/projections/omerc.js
@@ -218,7 +218,7 @@ export function inverse(p) {
 }
 
 // TODO: confirm all of these are the correct supported coordinates
-export var names = ["Hotine_Oblique_Mercator", "Hotine Oblique Mercator", "Hotine_Oblique_Mercator_Azimuth_Natural_Origin", "Hotine_Oblique_Mercator_Two_Point_Natural_Origin", "Hotine_Oblique_Mercator_Azimuth_Center", "omerc"];
+export var names = ["Hotine_Oblique_Mercator", "Hotine Oblique Mercator", "Hotine_Oblique_Mercator_Azimuth_Natural_Origin", "Hotine_Oblique_Mercator_Two_Point_Natural_Origin", "Hotine_Oblique_Mercator_Azimuth_Center", "Oblique_Mercator", "omerc"];
 export default {
   init: init,
   forward: forward,
--- a/lib/projections/omerc.js
+++ b/lib/projections/omerc.js
@@ -1,168 +1,224 @@
 import tsfnz from '../common/tsfnz';
 import adjust_lon from '../common/adjust_lon';
 import phi2z from '../common/phi2z';
-import {EPSLN, HALF_PI, FORTPI} from '../constants/values';
+import { D2R, EPSLN, HALF_PI, TWO_PI, FORTPI } from '../constants/values';
+
+var TOL = 1e-7;
 
 /* Initialize the Oblique Mercator  projection
     ------------------------------------------*/
-export function init() {
+export function init() {  
+  var con, com, cosph0, D, F, H, L, sinph0, p, J, gamma = 0,
+    gamma0, lamc = 0, lam1 = 0, lam2 = 0, phi1 = 0, phi2 = 0, alpha_c = 0, AB;
+  var no_off = 0;
+  
   this.no_off = this.no_off || false;
   this.no_rot = this.no_rot || false;
-
-  if (isNaN(this.k0)) {
-    this.k0 = 1;
+  
+  var alp = this.alpha !== 0;
+  var gam = (this.rectified_grid_angle && this.rectified_grid_angle !== 0) || false;
+  
+  if (alp) {
+    alpha_c = this.alpha;
   }
-  var sinlat = Math.sin(this.lat0);
-  var coslat = Math.cos(this.lat0);
-  var con = this.e * sinlat;
-
-  this.bl = Math.sqrt(1 + this.es / (1 - this.es) * Math.pow(coslat, 4));
-  this.al = this.a * this.bl * this.k0 * Math.sqrt(1 - this.es) / (1 - con * con);
-  var t0 = tsfnz(this.e, this.lat0, sinlat);
-  var dl = this.bl / coslat * Math.sqrt((1 - this.es) / (1 - con * con));
-  if (dl * dl < 1) {
-    dl = 1;
+  
+  if (gam) {
+    gamma = (this.rectified_grid_angle * D2R);
   }
-  var fl;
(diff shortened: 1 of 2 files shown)


dulldrums:
Here's a demo of something I threw together with the example from the issue here: https://github.com/proj4js/proj4js/issues/308#issue-392554286

It appears that the mapshaper-proj results match what is returned from proj4. (The version of proj4js used here is what is currently in this repository, not the implementation from my branch)

https://codesandbox.io/embed/r48wn6voq

calvinmetcalf:
ok so we need to test some of the failing tests to make sure that they are in fact real failures, when I created some of these tests I just plugged coordinates into this library and recorded their results and used those for the tests so that any refactoring we did wouldn't change the output, this doesn't necessarily mean they are correct though so if we plug the same projection and coordinates into a different library (or better yet 2) and they give a different result which is the same as the 'incorrect' error we get, we can probably just fix the test.

also the style is a bit wonky compared to other files but we don't have to worry about that until after we make sure it works.

lastly I just merged in #318 which fixes travis builds so you probably want to rebase off that

dulldrums:
@calvinmetcalf got it, thanks! I'll come up with a way to test all the values for omerc tests, and update them where it's necessary

dulldrums:
@calvinmetcalf Rebased and added a commit to fix those tests.

Also, added a test for the transformation @nethdeco described in their issue (#308).

The values I took for these tests are coming from proj4, but would love to have someone else verify that these values are accurate. I compared these values to values from QGIS, and noticed a slight difference for `CH1903 / LV03` ([EPSG:21781](http://epsg.io/21781)). It looks like this transformation isn't as accurate as the others being tested, but the difference seems greater than the others (e.g. [EPSG:2057](http://epsg.io/2057)). This fix returns virtually the same values as the previous implementation (acc: { xy: 9}), so it makes me think the issue is with inconsistencies in the definition of EPSG:21781.

Please let me know if there is anything else for this PR.

Thanks!

calvinmetcalf:
i wonder if this fixes the issues we had with [epsg 3468](http://epsg.io/3468) ?

dulldrums:
Hmm, it doesn't look like it is... Ill have a quick look to see what it could be

dulldrums:
It looks like it might have something to do with the `+no_uoff ` param being ignored. I compared this fix with mapshaper, and it looks like mapshaper is producing accurate results.

Looking at the Proj4 definition on http://epsg.io/3468, it appears the `+no_uoff ` param is there. I found this issue linked to Proj4 https://github.com/OSGeo/proj.4/issues/104#issue-79330678 . 

It looks like we need to do something like this somewhere: https://github.com/mbloch/mapshaper-proj/blob/master/src/wkt/projections/wkt_omerc.js#L8

calvinmetcalf:
I'm pretty sure it doesn't work if you if you use the WKT version of the proj definition as well, which doesn't have the no_uoff param at all

dulldrums:
> I'm pretty sure it doesn't work if you if you use the WKT version of the proj definition as well, which doesn't have the no_uoff param at all

Exactly. I'm trying to figure out how to assume the value of `no_uoff` if it isn't provided. Haven't been able to figure it out yet though

calvinmetcalf:
it looks like mapshaper uses the name to differentiate it, which should be available in there somewhere

lib/projections/omerc.js
```diff
@@ -181,8 +181,8 @@ export function inverse(p) {
   var u, v, Qp, Sp, Tp, Vp, Up;
   var coords = {};
   
-  p.x = (p.x * this.to_meter - this.x0) * (1.0 / this.a);
-  p.y = (p.y * this.to_meter - this.y0) * (1.0 / this.a);
+  p.x = (p.x - this.x0) * (1.0 / this.a);
```
calvinmetcalf: are we going to need to add some tests for non meter omerc projections ?

dulldrums: I would think so.

I'll get back to this PR sometime next week, got pulled into some other stuff.

dulldrums:
Just pushed a new commit up, a couple of things to note:

- This change fixes the issue with EPSG:3468. Tests have been added to reflect that.
- I couldn't find any valid proj4 string or WKT to test the projection "Hotine_Oblique_Mercator_Two_Point_Natural_Origin". Do you know any EPSG code that can be used to get the data to test this?
- The logic to check if the omerc is type A or B is dependent on what Projection string is provided. I did my best to make this logic as clear as possible.
- Tests were added for both Type A and Type B projections, using both Proj4 format and WKTs
- I updated the EPSG:2057 test with accurate values.
- I couldn't find any omerc projections that use ft, some tests for that could be added though

Let me know if there is anything else! Thanks!

dulldrums:
Just added a test for omerc type B using Foot_US as the unit

calvinmetcalf:
[epsg:54025](http://epsg.io/54025) should be Hotine_Oblique_Mercator_Two_Point_Natural_Origin, this is great!

dulldrums:
@calvinmetcalf I had seen EPSG:54025, but it doesn't seem to work on http://epsg.io/transform#s_srs=4326&t_srs=54025&x=1.0000000&y=1.0000000 or using mapshaper. I'm not sure if it is a problem with its definition.

childsd3:
Any update on getting this merged?

PR #319, opened 2019-04-23
Add "gstmerc" as a name for the gstmerc projection

CarlQLange:
I'm pretty sure `gstmerg` is a typo and it should be `gstmerc` instead. I've added `gstmerc` as a name but kept `gstmerg` to preserve backwards compatibility. ✌🏽


calvinmetcalf:
i think you are right,

Issue #195, opened 2016-06-30
Transforming EPSG:4978 geocent projection fails?


jwgmeligmeyling:
I am trying to transform points to and form the EPSG:4978 projection. I have found multiple definitions on the internet, but none of them seem to work. For example the empty definition: (source: http://spatialreference.org/ref/epsg/wgs-84-2/proj4js/)

``` js
proj4.defs("EPSG:4978", "");
```

Throws an error.

The defintion:

``` js
proj4.defs("EPSG:4978", "+proj=geocent +datum=WGS84 +units=m +no_defs");
```

Can be created, but transforming points fails with an error. For example:

``` js
proj4(proj4.defs("EPSG:4978"), proj4.defs("EPSG:4978"), [2000000, 600000]);
```

Then I found this issue on a mailing list: http://lists.osgeo.org/pipermail/metacrs/2010-September/000490.html

And this unanswered question on the GIS stackexchange: http://gis.stackexchange.com/questions/181929/proj4js-with-geocentric-crs

I am wondering, why is this failing? Is this a known issue? Am I using an older version of proj4js?

calvinmetcalf:
Thanks not a supported projection type

JWGmeligMeyling:
Aha, how can one impkement it?

calvinmetcalf:
#244 is probably a good guide, adding geocentric would probably be easier as there really aren't any parameters, might be a little tricky just because no projections currently require a z

codingdna2:
Can someone provide a simple example? I'm not able to use the workaround. I want to convert from ESPG:4326 to EPSG:4328 but I'm getting error with any system which uses +proj=geocent

ahocevar:
Fixed with #327.

PR #327, opened 2019-08-09
Added geocentric projections

lcalisto:
I have added support for geocentric projections, specially EPSG:4978 Addressing #195


crubier:
Nice ! We need that

liron14:
Hi, I really need this, when will it be merged into the master?

PR #337, opened 2019-11-29
Add support for Lambert Conformal Conic 1SP

hccampos:
This allows suporting, for example: https://epsg.io/3448

More information about the projection can be found here: https://proj.org/operations/projections/lcc.html


ahocevar (approved):
Thanks, @hccampos

PR #338, opened 2019-11-29
Respect datum params parsed from WKT

hccampos:
Proj4js was defaulting to an existing datum definition but now it only does so if no `TOWGS84` parameters are provided in the WKT.


ahocevar:
Thanks for your contribution, @hccampos. I resolved the conflict and created a new pull request #378 with your fix.

PR #339, opened 2019-11-29
omerc: prevent calculating Math.asin outside of the [-1, 1] range.

hccampos:
Because of floating point precision issues, some CRS definitions might end up causing `Math.asin` to be called with something like `1.000000000001` which results in `NaN`.


ahocevar:
With the new omerc implementation brought in by #350, this is no longer relevant.

Issue #341, opened 2020-01-01
LAEA very broken


pjsg:
If I add the following two tests, they both fail:

```
 },{
    code:'+proj=laea +lat_0=2 +lon_0=1 +x_0=0 +y_0=0 +R=6317000 +units=m +no_defs',
    ll:[2, 1],
    xy:[0, 0]
  },{
    code:'+proj=laea +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +R=6317000 +units=m +no_defs',
    ll:[19, 0],
    xy:[2085211.45, 0]
  },{
```
The `xy` value in the second case comes from proj4. It doesn't match the value output by proj4js (and the inverse doesn't bring it back to 19,0)

There are some simple fixes to make the first test case work, but the second is more concerning. Does anybody actually use LAEA?

pjsg:
Additionally I was confused by the fact that proj4js doesn't recognize +R for the `laea` projection. According to the PROJ.4 docs, it seems that +R overrides +a and +b and makes it spherical.

Issue #340, opened 2019-12-15
AEQD weirdness depending on precise shape of the earth


pjsg:
Maybe there is something that I don't understand, but the coordinate mapping for AEQD depends on the a & b parameters -- as to how coordinates are mapped when more than 90 degrees away in longitude.

```proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000.001 +b=6371000 +units=m', [91,0])
(2) [9896348.472919082, 6.059765740303575e-10]
proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000 +b=6371000 +units=m', [91,0])
(2) [10118738.324654846, 0]
```

This seems to indicate that the processing is very different if the earth is spherical / non-spherical.

For the same two shapes of the earth, Long 89, Lat 0 is essentially the same:

```proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000.001 +b=6371000 +units=m', [89,0])
(2) [9896348.472919082, 6.059765740303575e-10]
proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000 +b=6371000 +units=m', [89,0])
(2) [9896348.471365727, 0]
```

Something weird also happens at 0, 0

```proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000.001 +b=6371000 +units=m', [0,0])
(2) [0, 0]
proj4('+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +a=6371000 +b=6371000 +units=m', [0,0])
(2) [NaN, NaN]
```

pjsg:
There was a change in PROJ4 to fix this. The particular commit was:

https://github.com/OSGeo/PROJ/commit/de8d6779aaabe

PR #342, opened 2020-01-02
Fix problems with AEQD and LAEA especially at the origin point

pjsg:
This adds some tests for AEQD and LAEA (and one commented out test that doesn't work that I can't figure out how to fix).

The principal fixes are to make the origin point of the projection map correctly to 0,0

This fixes mostly #340 and #341


pjsg:
Bump -- I'd like to get these changes merged (I'm using these in openlayers).

PR #350, opened 2020-04-30
Omerc accuracy

### LINKED ISSUE #375: Error when converting EPSG:3168
I tried to convert with 
+proj=omerc +lat_0=4 +lonc=102.25 +alpha=323.0257905 +k=0.99984 +x_0=804670.24 +y_0=0 +no_uoff +gamma=323.1301023611111 +a=6377295.664 +b=6356094.667915204 +units=m +no_defs 
but result is way off.

However, one on epsg.io is correct.

childsd3:
Not really sure the best way to bring this into the main proj4js repository, but I added a couple changes that allow for the gamma and alpha to properly get resolved. 

The main fix with this is that when a gamma is defined as 0 in the proj string, `gam` should be true and should go through the rest of the equations.

This pull request builds on #317 

Fixes #375

### DIFF
--- a/lib/projString.js
+++ b/lib/projString.js
@@ -46,6 +46,9 @@ export default function(defData) {
     alpha: function(v) {
       self.alpha = parseFloat(v) * D2R;
     },
+    gamma: function(v) {
+      self.rectified_grid_angle = parseFloat(v);
+    },
     lonc: function(v) {
       self.longc = v * D2R;
     },
--- a/lib/projections/omerc.js
+++ b/lib/projections/omerc.js
@@ -218,7 +218,7 @@ export function inverse(p) {
 }
 
 // TODO: confirm all of these are the correct supported coordinates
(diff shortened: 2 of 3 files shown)


mariusk:
Is anybody looking into this MR? I think we're getting hit by it as well (gamma related inaccuracies).

ahocevar:
As already mentioned in #317, I'm basically fine with this pull request. However, it would be good to have a reference for the changed assertions in the tests. I'm also concerned about the one commented out test, which I'd also like to see an explanation for why it fails.

If anyone can show me that the new values are correct, and explain why that one test fails, I'll go ahead and merge this.

ahocevar (commented):
I just verified the tests, and this pull request does not break any existing tests. In addition, it fixes the newly reported issue #375. The code looks good, so I am going to merge this. Thanks @childsd3 and @dulldrums for these improvements!

childsd3:
Thanks @dulldrums, @ahocevar, and @unknowdata!

Issue #62, opened 2013-09-24
add support to Proj.4 nadgrids parameter (grid based datum adjustments)


exPangea:
Proj.4 nadgrids parameter cannot be passed to proj4.js (at least when using it in client browser) 

This is necessary for grid based datum adjustments, as described in the official Proj.4 documentation: 
https://proj4.org/usage/transformation.html?highlight=nadgrids#grid-based-datum-adjustments

Example datum shift grids are available at https://proj4.org/resource_files.html#transformation-grids

If possible, it would be desirable to have the choice of loading grid files from server, client (local filesystem) and/or remote urls.

andreasplesch:
Looking at earlier versions, I found this ticket which implemented support for nadgrids:

https://trac.osgeo.org/proj4js/ticket/81#no1

Was this not fully compatible with new proj4js or not functional ?

calvinmetcalf:
@andreasplesch it was started but never finished so it was removed due to it not functioning

andreasplesch:
Since it was a port from proj4, this looked like a good start. Oh well, perhaps the added accuracy is not worth the effort in most use cases.

calvinmetcalf:
If somebody wanted to add the functionality that would be the place to
start

andreasplesch:
ok.

fakharkhalid:
Are there any plans in the proj4js roadmap to address this issue?

ahocevar:
@fakharkhalid If anyone wants to contribute that feature, it would be a welcome pull request. The only thing to keep in mind is that grids should not be included in the build, because otherwise the library size would get too big for browser use.

geoffsmith:
I've made a start on implementing this: [https://github.com/geoffsmith/proj4js/tree/nadgrids](https://github.com/geoffsmith/proj4js/tree/nadgrids)

That branch includes a rough working version, though it will only work for node right now. My plan over the next few days is to figure out how to get it working for the browser, clean it up, then make a pull request. I'm quite new to projections, so I had to rely heavily on the old implementation linked to above. I've cleaned it up a little and made a few changes to get it to work (that I think might have been typos). I've tested it against the UK Ordnance Survey nadgrid file, they handily  [provide a set of test points]( https://www.ordnancesurvey.co.uk/business-government/tools-support/os-net/for-developers) that I've used to verify the code works.

[...]

PR #363, opened 2020-10-01
Adding support for nadgrids

geoffsmith:
I've added basic support for nadgrids and NTv2 grid files. I'm a projection novice, so this relied heavily on the patch linked to in the original issue: https://github.com/proj4js/proj4js/issues/62. The patch showed me where to apply the grid shifts and provided the algorithm for looking up the correct shift, though I've tried to clean up the code a little, removed some parts I didn't see how they were contributing (though happy to be corrected on those parts) and made some changes where the logic wasn't right (and it broke the tests). On top of this I implemented a basic NTv2 file reader.

Since there are many different ways a consumer of the API might want to load the binary NTv2 file, my implementation doesn't attempt to read in the binary, but expects the caller to provide a pre-loaded ArrayBuffer with the contents of their NTv2 file. My thinking is that trying to support all the ways that people will need to access the file would quickly get out of hand (e.g. in node, in the browser, fetch vs XMLHttpRequest, auth headers, etc, etc). So I've exposed a new function in the index: `nadgrid(key, buffer)` which will parse the ArrayBuffer and associate it with the given key. The user will then specify the key in the proj4 string, e.g. `+nadgrids=@key,null`.

I originally tested this using the [Ordnance Survey NTv2 file](https://www.ordnancesurvey.co.uk/business-government/tools-support/os-net/for-developers), which also provides a set of test points, which was great for building confidence in the implementation. However, there's no license specified with those files, so I've replaced those tests with some that are based on a downsampled ntv2 grid that I found in the OSGeo/proj repo. Again, trying to load a binary file across different browsers, node etc is a pain, so in the end I base64 encoded the file to make it easier to load into the tests. If anyone can think of a nicer way of doing this, I'm more than happy to change it.


ahocevar:
Thanks a lot for your work on this @geoffsmith! I have not yet found time to give this the proper review it deserves, so I'm asking you for a little patience. I'll make sure to review this, but it may be a few more weeks before I get to it.

geoffsmith:
@ahocevar thanks for letting me know - looking forward to hearing what you think :)

olivierchatry:
Started using this branch, I think I found a bug :
in Proj.js, l30, the `nadgrids` returned by `match` is ignored, so "default" nadgrids defined are not loaded.

I added that : `json.nadgrids  =  datumDef.nadgrids ? datumDef.nadgrids : json.nadgrids`

geoffsmith:
@olivierchatry thanks for trying out this change! I'll have a look...

olivierchatry:
I'm trying to see if I can add subgrids support. I'm guessing it would be desirable to have some kind of hashing, or binary tree maybe for fast lookup ?

olivierchatry:
Hum, I'm trying to do that : 
```
proj4.defs("EPSG:32066", '+proj=tmerc +lat_0=0 +lon_0=-87 +k=0.9996 +x_0=500000.001016002 +y_0=0 +datum=NAD27 +units=us-ft +no_defs')

const proj = proj4('EPSG:4326', 'EPSG:32066')

console.log(proj.forward([-87.9336317, 28.57421]))
```

And this seems to fails quite hard. Trying to debug where it fails.

olivierchatry:
Yup, I think there is an issue with min max testing : 
```
         if (minY > input.y || minX > input.x || maxY < input.y || maxX < input.x ) {
```

should be :
```
        if (minY > input.y && minX > input.x && maxY < input.y && maxX < input.x ) {
```

geoffsmith:
@olivierchatry Did you load the nadgrid file into memory using something like this?

```javascript
const buffer = fs.readFileSync('ntv2_0.gsb').buffer
proj4.nadgrid('ntv2_0.gsb', buffer);
```

olivierchatry:
No, I'm in chrome. Loading is fine ( I added loading of multiple grid and support for it in the computation ). Fixed the bounding box check. My issue was the transfo I'm using use conus. Not sure where I can find the glb file for it.

geoffsmith:
Is it one of these maybe? https://www.ngs.noaa.gov/GEOID/GEOID12B/GEOID12B_CONUS.shtml

geoffsmith:
@olivierchatry I've had a little dig into this and I think I've found the conus files. [Here's a page](http://www.teledynecaris.com/docs/4.4.11/hips%20and%20sips/index.html#page/CARIS%20HIPS%20and%20SIPS%20Help/CARISFiles_CRS.32.5.html) describing some grid transforms: . By googling those file names, [I've found some copies of them](https://ashgps.com/2/FAST%20Survey/Software/Utilities/Grids/USA%20-%20Nadcon/NADConv/index.htm).

It looks to me like the conus grid transforms use a different file format NADCON, whereas I've only added support for NTSv2.

Coming back to the subgrids. The thing that needs changing is to follow the algorithm define in [this document](https://web.archive.org/web/20140127204822if_/http://www.mgs.gov.on.ca:80/stdprodconsume/groups/content/@mgs/@iandit/documents/resourcelist/stel02_047447.pdf) under the "Subroutine FGRID" section. Basically find the top-level grid that includes the point to transform, then recursively check any child grids to see if they contain the point. Pretty simple to add I think, but might not help in your case since you're blocked on the NADCON file format.

olivierchatry:
Yup, I decided to create a C++ micro service and use the fullblow proj instead. In the long run it will be easier for my use case.

ahocevar (approved):
Excellent work, @geoffsmith. This is a huge improvement to the library, and a frequently requested new feature.

I'm going to create a small follow-up pull request to simplify the tests (i.e. use the gird file directly) and provide a bit more detail in the README.

PR #386, opened 2021-06-08
Update Hermannskogel and Potsdam to 7 parameter Helmert transform

### LINKED ISSUE #385: Hermannskogel and Potsdam datums are using 3-parameter instead of 7-parameter Helmert transformation
We see a larger error of > 100 m when transforming between WGS 84 and EPSG:31256. These two datums are using the three-parameter Helmert transformation in Proj4JS. In Proj4 these have been changed many years ago to the seven parameter version.

https://github.com/OSGeo/PROJ/issues/207
https://github.com/OSGeo/PROJ/issues/115

Would you be willing to accept a pull request that changes these two?
Are there maybe more definitions that are outdated?

rhuitl:
This PR updates the Helmert transform parameters to match what Proj is doing https://github.com/OSGeo/PROJ/blob/master/src/datums.cpp#L52 (at least before they updated Potsdam once more to use the Beta 2007 grid file). Also all the other datums that both projects have in common should use the same parameters now.

Fixes #385.

### DIFF
--- a/lib/constants/Datum.js
+++ b/lib/constants/Datum.js
@@ -31,7 +31,7 @@ exports.nad27 = {
 };
 
 exports.potsdam = {
-  towgs84: "606.0,23.0,413.0",
+  towgs84: "598.1,73.7,418.2,0.202,0.045,-2.455,6.7",
   ellipse: "bessel",
   datumName: "Potsdam Rauenberg 1950 DHDN"
 };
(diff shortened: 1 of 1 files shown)


rhuitl:
This PR updates the Helmert transform parameters to match what Proj is doing https://github.com/OSGeo/PROJ/blob/master/src/datums.cpp#L52 (at least before they updated Potsdam once more to use the Beta 2007 grid file).

rhuitl:
Oops, true that commit shouldn't be there. Fixed.

ahocevar (approved):
Thanks, @rhuitl!

PR #396, opened 2021-11-30
Parse +over param to disable longitude wrapping

risoo10:
Implementation of **over** parameter as described in official [PROJ documentation](https://proj.org/usage/projections.html#cartographic-projection):

`+over` - Allow longitude output outside -180 to 180 range, disables wrapping


ahocevar (commented):
Thanks @risoo10 for this contribution, and sorry for letting you wait for a review for so long. The code looks good, but for such a library-wide change, it would be great if you could add a few tests.

mclaeysb:
This would be very useful for us. Any chance in this PR making it through?

PR #405, opened 2022-03-04
Add 'geos' projection

### LINKED ISSUE #404: Advice adding 'geos' projection
I'm a developer on the GeoSphere (https://geosphere.ssec.wisc.edu/) site where we have created a modified version of proj4js that includes the geostationary projection calculations used in the PROJ C library. The developer who added the formulas is familiar with javascript but wasn't familiar with how to contribute to an open source library so the code was kept in our internal copy. I'd like to contribute this code to the library, but before I start cleaning up the code (javascript is not my primary language) I wanted to make sure this is something the maintainers are interested in or at least would accept. If so, what documentation exists for how to contribute to the project? I see the Developing section in the README [here](https://github.com/proj4js/proj4js/blob/master/README.md#developing) which will be very helpful. If there is anything else out there, let me know. If there are any gotchas you foresee I'm all ears for those tips too.

djhoese:
Closes #404

As described in #404, this is an implementation of the 'geos' projection based on the PROJ C library and copied from a custom internal implementation my group was using. The original javascript for the geos projection was hacked into the compiled source for proj4js so I copied it over and did the necessary changes to make it look like the other projection modules here. Keep in mind while reviewing that javascript is not my thing so feel free to comment on style, question why something is done, etc. That said, I'm very familiar with the pyproj library (Python wrapper around PROJ C) and the general concept of projections.

I've only added one test so far and it fails locally so I'm making this PR now to get some further advice. My questions are:

1. What should forward/inverse return when the calculation is invalid? For PROJ/pyproj this would normally return infinity I think. It looks like `merc` uses `null`, but was wondering if the `p` object should have `p.x` and `p.y` set to `null`?
2. Is there an easy way to only run my tests?
3. Is there an easy way to insert a debugger or even show the output of some `console.debug` calls or something and see the output when running the tests? Again, not familiar with common knowledge for JS development.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,7 +31,8 @@ var projs = [
   'qsc',
   'robin',
   'geocent',
-  'tpers'
+  'tpers',
+  'geos'
 ];
 module.exports = function (grunt) {
   grunt.initConfig({
--- a/lib/Proj.js
+++ b/lib/Proj.js
@@ -48,7 +48,7 @@ function Projection(srsCode,callback) {
   extend(this, json); // transfer everything over from the projection because we don't know what we'll need
   extend(this, ourProj); // transfer all the methods from the projection
 
-  // copy the 4 things over we calulated in deriveConstants.sphere
+  // copy the 4 things over we calculated in deriveConstants.sphere
   this.a = sphere_.a;
   this.b = sphere_.b;
   this.rf = sphere_.rf;
--- a/lib/includedProjections.js
+++ b/lib/includedProjections.js
@@ -25,6 +25,7 @@ import qsc from './projections/qsc';
 import robin from './projections/robin';
 import geocent from './projections/geocent';
 import tpers from './projections/tpers';
+import geos from './projections/geos';
 
 var projs = [
   tmerc,
@@ -53,7 +54,8 @@ var projs = [
   qsc,
   robin,
   geocent,
-  tpers
+  tpers,
+  geos,
 ];
 
 export default function (proj4) {
--- a/lib/projections/geos.js
+++ b/lib/projections/geos.js
@@ -1,0 +1,146 @@
+import hypot from '../common/hypot';
+
+export function init() {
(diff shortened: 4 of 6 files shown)


djhoese:
Ok I see where `console.debug` messages show up in the `npm test` output. Now I just need to figure out why I'm not getting the expected results. My test case should be a valid test.

ahocevar:
Thanks for tackling this, @djhoese!

> What should forward/inverse return when the calculation is invalid? For PROJ/pyproj this would normally return infinity I think. It looks like merc uses null, but was wondering if the p object should have p.x and p.y set to null?

`p.x` and `p.y` should be set to `NaN`.

> Is there an easy way to only run my tests?

When you add tests to `test/test.js`, you can change a `describe()` or `it()` to `describe.only()` or `it.only()` to only run that test.

> Is there an easy way to insert a debugger or even show the output of some console.debug calls or something and see the output when running the tests? Again, not familiar with common knowledge for JS development.

Looks like you figured out the console part already. For debugging with breakpoints et all, run

    npm run build
    node --inspect-brk node_modules/mocha/bin/_mocha test/test.js

Then open a Chrome browser, navigate to chrome://inspect, and click your test under "Remote target". To set a breakpoint programmatically, add
```js
    debugger // jshint ignore:line
```
in your code.

djhoese:
Thanks @ahocevar! Luckily I've also converted this geos logic to OpenGL shaders so the issues were familiar. This PR is ready for review and now has 4 test cases.

djhoese:
No rush on this, but wanted to make sure my updates didn't get lost on a weekend. I also wanted to comment to ask what your feelings and schedule are for releases? Thanks again for your help and the project.

ahocevar:
@djhoese Thanks for your contribution! Unfortunately I did not find time yet to give this a review. I'll review it as soon as I can, and I'll also publish a release when your changes are in.

ahocevar (approved):
Thanks @djhoese for this valuable contribution!

Issue #408, opened 2022-05-17
The 'qsfnz' function expects 2 arguments, but 3 were provided.


Oleksiy99:
Hello, I use the proj4-src.js file in my project and I run code analysis on my project. 
The analyzer says there are errors with proj4-src.js.
The issue exists when running the 'qsfnz' function in the proj4-src.js file. The 'qsfnz' function has two parameters, but in some places in this file you are running this function with three parameters. There are four places where you run this function with three parameters. 
In my project, I fixed this bug by removing the third parameter during the run of the 'qsfnz' function.
![bandicam 2022-05-17 17-13-45-118](https://user-images.githubusercontent.com/59996298/168832552-76fda133-f53e-4729-b6cd-5d7f0191b99f.jpg)
![bandicam 2022-05-17 17-13-52-319](https://user-images.githubusercontent.com/59996298/168832563-ce5c0f74-0894-4618-bdb8-500f92f1ec71.jpg)

Please consider this issue.

ahocevar:
Thanks for reporting this, @Oleksiy99. Do you think you'd be able to provide a pull request with your suggested fix?

Note that you wouldn't make this change in `proj4-src.js`, but in the respective source files in the `lib/` folder.

PR #410, opened 2022-05-18
Remove the third parameter for running the qsfnz function

Oleksiy99:
Fixes https://github.com/proj4js/proj4js/issues/408


ahocevar (approved):
Thanks, @Oleksiy99

Issue #412, opened 2022-08-24
Inaccurate results for EPSG:31370 to EPSG:4326


pieterprovoost:
I'm getting inaccurate results when converting from EPSG:31370 to EPSG:4326 with version 2.8.0. See fiddle [here](https://jsfiddle.net/5ukmdnt9/1/) which returns `3.5330645898513002, 50.97977690061455` and compare with `gdaltransform`:

```
% gdaltransform -s_srs EPSG:31370 -t_srs EPSG:4326
Enter X Y [Z [T]] values separated by space, and press Return.
91371.198492 185861.916314
3.53387845650625 50.9800414640608 0
```

The result is the same for proj4 and WKT, so not related to #379. The difference for my point is around 60 meters.

ahocevar:
Are you sure that gdaltransform uses the same transform as your fiddle? https://epsg.io/31370 lists several different transforms.

KaloNK:
If you check the transformation at epsg.io - it shows the correct results 3.5338783 and 50.9800413:
https://epsg.io/transform#s_srs=31370&t_srs=4326&ops=15929&x=91371.1984920&y=185861.9163140
Had the same problem with other EPSG containing transformations. The problem lies in the transformation functions geocentricToWgs84 and  geocentricFromWgs84. I have replaced them with my own function:

```
    function TransformGeocentric(FromWGS84, p, datum_type, datum_params) {
      // FromWGS84 = 1 if to WGS FromWGS84 = -1
      if (datum_type === PJD_3PARAM) {
        return {
          x: p.x - FromWGS84 * datum_params[0],
          y: p.y - FromWGS84 * datum_params[1],
          z: p.z - FromWGS84 * datum_params[2],
        };
      } else if (datum_type === PJD_7PARAM) {
          var M = datum_params[6];
          if ( FromWGS84 > 0 ) M = 1.0 / M;
          return {
            x: ((M * (p.x - FromWGS84 * datum_params[5] * p.y + FromWGS84 * datum_params[4] * p.z)) - FromWGS84 * datum_params[0]),
            y: ((M * (FromWGS84 * datum_params[5] * p.x + p.y - FromWGS84 * datum_params[3] * p.z)) - FromWGS84 * datum_params[1]),
            z: ((M * (-FromWGS84 * datum_params[4] * p.x + FromWGS84 * datum_params[3] * p.y + p.z)) - FromWGS84 * datum_params[2])
          };
      }
    }

```

[...]

PR #418, opened 2022-11-26
Fix transformations

KaloNK:
geocentricToWgs84 and geocentricFromWgs84 replaced with a single function

### DIFF
--- a/lib/datumUtils.js
+++ b/lib/datumUtils.js
@@ -170,76 +170,22 @@ export function geocentricToGeodetic(p, es, a, b) {
 } // cs_geocentric_to_geodetic()
 
 /****************************************************************/
-// pj_geocentic_to_wgs84( p )
-//  p = point to transform in geocentric coordinates (x,y,z)
 
-
-/** point object, nothing fancy, just allows values to be
-    passed back and forth by reference rather than by value.
-    Other point classes may be used as long as they have
-    x and y properties, which will get modified in the transform method.
-*/
-export function geocentricToWgs84(p, datum_type, datum_params) {
-
+export function TransformGeocentric(FromWGS84, p, datum_type, datum_params) {
+  // FromWGS84 = 1 if to WGS FromWGS84 = -1
   if (datum_type === PJD_3PARAM) {
-    // if( x[io] === HUGE_VAL )
-    //    continue;
     return {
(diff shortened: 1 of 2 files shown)


ahocevar (changes requested):
Thanks for trying to fix this, @KaloNK. 304 tests are failing with this change. But before you spend more time on this, did you follow my advice in #412 to check if you were using the same transform in proj4 as in your reference example?

KaloNK:
I am working with EPSG:28405 (  https://epsg.io/28405 ) and using exactly the string provided for poj4js on that page:
`proj4.defs("EPSG:28405","+proj=tmerc +lat_0=0 +lon_0=27 +k=1 +x_0=5500000 +y_0=0 +ellps=krass +towgs84=23.57,-140.95,-79.8,0,-0.35,-0.79,-0.22 +units=m +no_defs +type=crs");`

Without my change the results are shifted 5 to 20 meters from the same results at epsg.io - https://epsg.io/transform#s_srs=4326&t_srs=28405&x=27.9217725&y=43.1873279

After the change, the results are identical.

ahocevar:
So far so good. Now in order to be able to accept this pull request, we need to know whether the 304 failing tests make wrong assumptions (some results are 200 m off after your change), or how to proceed.

KaloNK:
To manually check (on epsg.io) all 304 will take a lot of time, but maybe just those that are too different.
If 10 of 10 results are confirmed on epsg.io it should be enough i think.

ahocevar:
I can confirm your findings in the two cases you provide, but I find it hard to believe we've been doing the Helmert transformation wrong since forever.

i checked some of the tests, and I was able to confirm that our assumptions for EPSG:31370 were indeed wrong. But all the others I checked were correct before your change, and are failing with your change.

To make more sense out of your change, would you be able to provide a citation for your code or the algorithm that you used?

KaloNK:
just followed https://en.wikipedia.org/wiki/Helmert_transformation (Application):
"for each single parameter of the coordinate"  and "For the reverse transformation, each element is multiplied by −1."

ahocevar:
Your're using the wrong sign for the matrix parameters when transforming to WGS84, and not applying the scale for the input points when transforming back. Like I wrote in #412, there is no need for this pull request, because nothing is broken.

Issue #460, opened 2024-01-17
The inverse method of the Ellipsoidal Azimuthal Equidistant projection returns incorrect results


EkkoScholtens:
The inverse method of the Ellipsoidal Azimuthal Equidistant gives incorrect results for inputs in the ranges:
* Latitude between [-90, 0] degrees
* Longitude between the ranges [-180, -90] degrees and [90, 180] degrees

I've tested these values for the with the following code:
```js
    proj4.defs(
        "aeqd",
        '+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84 +units=m'
    );

    let input = [0,-52];
    let aeqdCoord = proj4("EPSG:4326", "aeqd", input);
    let wsg84Coord = proj4("aeqd", "EPSG:4326", aeqdCoord);
    console.log(input); // [0, -52]
    console.log(aeqdCoord); // [7.058137128712722e-10, -5763406.341834129]
    console.log(wsg84Coord); // [8.923286760849614e-15, 127.9924279990819] expected something near [0, -52]

    let input2 = [145,0];
    let aeqdCoord2 = proj4("EPSG:4326", "aeqd", input2);
    let wsg84Coord2 = proj4("aeqd", "EPSG:4326", aeqdCoord2);
    console.log(input2); // [145,0]
    console.log(aeqdCoord2); // [3896182.177764574, 2.385723516447175e-10]
    console.log(wsg84Coord2); // [34.99999999999999, 2.0258715108875234e-15] expected something near [145,0]

```

I believe to fix the issue with the latitude the following line:
https://github.com/proj4js/proj4js/blob/3c3e1f46c52d3291992f7b5c2d74f42a4a231968/lib/projections/aeqd.js#L193
needs to be: `lat = Math.atan((sinpsi - this.es * F * this.sin_p12) * Math.tan(psi) / (sinpsi * (1 - this.es)));`

EkkoScholtens:
I can make the pull request for that issue. But i don't know what the issue is with the longitude. I believe this code is implemented from the book "Map Projections - A Working Manual" By JohnP. Snyder right?

ahocevar:
[...]

PR #461, opened 2024-01-19
Fix for the incorrect results when using the inverse method of the Ellipsoidal Azimuthal Equidistant projection.

EkkoScholtens:
This fixes the issue mentioned in the issue #460 where the inverse method of the Ellipsoidal Azimuthal Equidistant gives incorrect results for inputs in the ranges:

- Latitude between [-90, 0] degrees
- Longitude between the ranges [-180, -90] degrees and [90, 180] degrees

I've fixed it using the [geographiclib](https://www.npmjs.com/package/geographiclib-geodesic) library used in the c++ implementation of [PROJ](https://github.com/OSGeo/PROJ) and used in [mapshaper-proj](https://github.com/mbloch/mapshaper-proj/tree/master).


ahocevar:
Thanks for the fix! The code changes look good, but I'm wondering why you've made a significant change to the test stack in the same commit. 

I appreciate changes that update our tooling, but it would be better to have them done in a dedicated pull request. Would you be able to change this pull request so it only fixes what the title says, and make another pull request with the tooling changes, including the removal of now unused packages and showing a way to run the tests in the browser.

ahocevar:
The bundle (`dist/proj.js` and `dist/proj-src.js`) has become invalid with this pull request, because the `geographiclib-geodesic` import is not handled correctly by the build tools.

EkkoScholtens:
> The code changes look good, but I'm wondering why you've made a significant change to the test stack in the same commit.

Geographiclib uses some language features that are not suported by the current versions of the build tools. I did some work on updating the test stack and build system, i will put them in to a seperate pull request. But that does mean that that needs to be merged first.

ahocevar:
With your recent changes, grunt never completes. It gets stuck with "waiting forever" after creating a server on port 8080. But this is minor, because it should be fixed with #462.

My doubt, however, is if it makes sense to pull in an external library that is used for just a single projection. The increase in bundle size is about 25%. Because proj4js is used in browser applications a lot, I'm wondering if there wouldn't be an alternative way to fix aeqd, without pulling in an external library that isn't used anywhere else in proj4js.

EkkoScholtens:
> My doubt, however, is if it makes sense to pull in an external library that is used for just a single projection. The increase in bundle size is about 25%. Because proj4js is used in browser applications a lot, I'm wondering if there wouldn't be an alternative way to fix aeqd, without pulling in an external library that isn't used anywhere else in proj4js.

Maybe there is but considering that both [PROJ](https://github.com/OSGeo/PROJ) and [mapshaper-proj](https://github.com/mbloch/mapshaper-proj/tree/master) went this way, makes me think that this is the best way. 

PROJ 4.9 did use the same method as you implemented here, but from version 5 on they included geographiclib for only this projection.

ahocevar:
I'm leaving this open for now. If I find time, I'll dig into the geographiclib source code and see what they do different from the current implementation.

ahocevar:
@EkkoScholtens I looked up the history of aeqd, and foud the PROJ issue that suggests the use of geographiclib https://github.com/OSGeo/PROJ/issues/246. So yes, your changes make sense - to be in line with other implementations, and for increased accuracy.

Also, your changes to modernize the build process are appreciated. Given the increased bundle size, I'm inclined to change the new build system (#462) and library structure to leverage dynamic imports based on projection definition, and to make the library also work without using its pre-built bundle.

lib/projections/aeqd.js
```diff
@@ -177,26 +166,17 @@ export function inverse(p) {
...
-      lat = Math.atan2((sinpsi - this.es * F * this.sin_p12) * Math.tan(psi), sinpsi * (1 - this.es));
-      p.x = lon;
-      p.y = lat;
+      lat1 = this.lat0 / D2R;
+      lon1 = this.long0 / D2R;
+      azi1 = Math.atan2(p.x, p.y) / D2R;
+      s12 = Math.sqrt(p.x * p.x + p.y * p.y);
+      vars = this.g.Direct(lat1, lon1, azi1, s12, this.g.STANDARD);
+
+      p.x = (vars.lon2 * D2R) - this.long0;
```
danielschilling-ml, line 175: Oblique projections where `+lat_0` and `+lon_0` are at weird angles don't work with this subtraction of `this.long0`.  `g.Direct` already takes care of this rotation. It should be this:

```
p.x = vars.lon2 * D2R;
```

test/testData.js
```diff
@@ -242,14 +242,30 @@ var testPoints = [
...
+  },{
+    code:'+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
+    ll:[145, 0],
+    xy:[ 16141326.16502467, 0]
+  },{
+    code:'+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
+    ll:[-145, 0],
+    xy:[-16141326.16502467, 0]
   },{
 //    code:'+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
```
danielschilling-ml, line 270: This commented-out test now works with your changes!

test/testData.js
```diff
@@ -242,14 +242,30 @@ var testPoints = [
...
+    ll:[0, -52],
+    xy:[0, -5763343.550010418 ]
+  },{
+    code:'+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
+    ll:[145, 0],
+    xy:[ 16141326.16502467, 0]
+  },{
+    code:'+proj=aeqd +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
+    ll:[-145, 0],
+    xy:[-16141326.16502467, 0]
```
danielschilling-ml, line 268: It would be good if this had an oblique test case where `+lat_0` and `+lon_0` are not equal to zero, like:

```js
  {
    code:'+proj=aeqd +lat_0=83.6625 +lon_0=-29.8333 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
    ll:[150.1667, 87.38418697931058],
    xy:[0, 1000000]
  }
```

danielschilling-ml (commented):
Your PR was really helpful to me! Thank you so much. Here are some suggestions to make it even better regarding problems I ran into with it.

ahocevar:
If this pull request can be completed to also fix the cases that @danielschilling-ml brought up, I'd be willing to accept the fact that `geographiclib-geodesic` is required. @EkkoScholtens Apologies for not having looked at #462 for over a year!

EkkoScholtens:
@ahocevar  No problem, I've made the changes that @danielschilling-ml brought up and added the tests.

ahocevar:
Thanks, @EkkoScholtens!

@danielschilling-ml Do you Are you happy with this now?

ahocevar (approved):

ahocevar:
With no more feedback from @danielschilling-ml, I have merged this now. Thanks, @EkkoScholtens!

PR #462, opened 2024-01-22
Update build system

EkkoScholtens:
I've updated the build and tests system to newer and non-deprecated packages. This is needed for working with third party packages, like Geographiclib that is needed by #461. 

Changes made:
## Moved rollup to separate npm task
I've updated the version of rollup to the latest version and created a separate npm task for it. This reduces the dependency on Grunt. The plugins for rollup have been updated to the latest versions too.

## Jshint and Uglify 
JsHint has been replaced by esLint, the config has been ported over to esLint. This did reveil some not correctly formatted files and two unused variables. Uglyfy has been replaced terser, which is a minifier for rollup.

Both have been moved from a Grunt task to a rollup task becode the grunt tasks where no longer maintained. 

## Testing 
Mocha is still used but I've updated the version to the latest. However mocha_phantomjs have been replaced with a custom test runner and puppeteer, since phantomjs has been deprecated and does not support the latest versions of ES, however there is no drop in replacement for it. So i needed to write a script that starts a file server and tells puppeteer to load the test page. The test page now returns a json object that contains the test results which gets parsed by the puppeteer script to pass or fail the run.


ahocevar:
Thanks a lot for your contributions, @EkkoScholtens. Please be patient if it takes a while before I get to reviewing them, because I have very limited time for this voluntary job of maintaining proj4js.

ahocevar:
Thanks a lot for your efforts, @EkkoScholtens. I took the liberty to commit final tweaks.

PR #477, opened 2024-09-15
Added named export `proj4` support to index.js

### LINKED ISSUE #476: ESM vs CommonJS exports inconsistency
By default, the `dist/proj4-src.js` file generates an 'exports' field using the grunt build, which does not provide a default export,

meanwhile, the module `lib/index.js` provides only a default export.

In environments where there is a mismatch between library usage (for example, we experience this when building our app as ESM, which uses the `lib/index.js`, but test using jest which only supports commonjs transformation and as such enters the `dist/proj4-src.js` file) the simple:
```ts
import proj4 from 'proj4';
```

breaks, as the tests which use commonjs transformation use the dist file, which throws the error that proj4 is undefined (there is no default export).

the solution is to support the named export proj4, such that both:

```ts
import * as proj4 from 'proj4';
```

will work with both `module` and `main` entry points.

Beilinson:
Fixes #476 

I hope this can be merged soon, as this issue is actively breaking unit tests in our GIS projects, and additionally is good practice to not have only default exports

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "proj4",
-  "version": "2.12.3-alpha",
+  "version": "2.12.2-alpha",
   "description": "Proj4js is a JavaScript library to transform point coordinates from one coordinate system to another, including datum transformations.",
   "homepage": "https://proj4js.github.io/proj4js/",
   "main": "dist/proj4-src.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "proj4",
-  "version": "2.12.2-alpha",
+  "version": "2.12.3-alpha",
   "description": "Proj4js is a JavaScript library to transform point coordinates from one coordinate system to another, including datum transformations.",
   "homepage": "https://proj4js.github.io/proj4js/",
   "main": "dist/proj4-src.js",
--- a/lib/index.js
+++ b/lib/index.js
@@ -19,4 +19,5 @@ proj4.transform = transform;
 proj4.mgrs = mgrs;
 proj4.version = '__VERSION__';
 includedProjections(proj4);
+export { proj4 };
 export default proj4;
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "proj4",
-  "version": "2.12.2-alpha",
+  "version": "2.12.3-alpha",
```
ahocevar, line 3: Please remove this change - that will be done as part of release management.

ahocevar (changes requested):

ahocevar (approved):

ahocevar (changes requested):
As you can see, this breaks the build.

Beilinson:
Yep, looking into it locally

Beilinson:
From my local tests this works, and also manages to fix the problem with jest in a different way, by exposing shimming the .default property that some parsers use when transforming ESM code to CJS.

I noticed that this is the same change as from #263 which is 7 years old, any reason that wasn't ever merged?

ahocevar:
It's really about time to update the build system - along the lines of #462, but getting rid of grunt entirely.

In the meantime, I think a safer change would be to create a separate entry point file for rollup (e.g. `./rollup/index.js`), which imports `proj4` from `../lib/index.js` and applies the hack you're proposing. Then you could point rollup to `rollup/index.js` instead of `lib/index.js` in the `Gruntfile`.

Then when updating the build system, we just get rid of `rollup/index.js`, and everything should work without this hack.

Beilinson:
having done some more tests, I realized that what this PR achieves can also be done by enabling [esModuleInterop](https://www.typescriptlang.org/tsconfig/esModuleInterop.html) which does exactly what this fixes ascribes. Closing this, if you would like any help with continuing #462 (for example, upgrading to jest as well, which supercedes mocha) I would be glad to help

ahocevar:
@Beilinson I'd prefer vitest over jest, but help is always appreciated. Thanks!

PR #479, opened 2024-10-09
CompoundCrs: adding the sub crs to defs

ftoromanoff:
Following the PR https://github.com/proj4js/wkt-parser/pull/30.
I suggest in the case of a Compound crs to add all the sub crs to defs. The user then can decide to only use the crs he prefers as Compouds crs in not yet supported by proj4.

### DIFF
--- a/lib/defs.js
+++ b/lib/defs.js
@@ -12,7 +12,14 @@ function defs(name) {
         defs[name] = parseProj(arguments[1]);
       }
       else {
-        defs[name] = wkt(arguments[1]);
+        const proj = wkt(arguments[1]);
+        if (['COMPOUNDCRS', 'COMPD_CS'].includes(proj.type)) {
+          Object.keys(proj).forEach(key => {
+            if (Array.isArray(proj[key]))
+              defs[proj[key].title] = proj[key];
+          })
+        }
+        defs[name] = proj;
       }
     } else {
       defs[name] = def;


ahocevar:
Looking closer at it, this pull request would create projection definitions that the user would not expect. The proj4 API assigns a single identifier to a single proj definition:
```js
proj4.defs('myid', 'mydef');
```
What you're proposing here will cause other defs than just the `myid` one to be created with this single API call.

The changes you suggest in wkt-parser make sense to some extent, but I think we better work out a good API there rather than trying to add something here that makes the API work in unexpected ways.

ftoromanoff:
Currently there is no way to get an identifier from a wkt (or projString). To parse a wkt I have to use an 'id' as you pointed (ie (proj4.defs(myid, mywkt)), then i will need myid to get the proj4 associated properties.
My approche was to call defs( unknown', mywkt) then to extract an id from the proj4.defs(''unknown') and then use it as an alias proj4.defs(proj4.defs(''unknown').id, proj4.defs(''unknown')).

An other solution might be to be able to call proj4.defs directly on a wkt that will return an id.
ie proj4(mywkt) => mywkt.id.

In this case it might be good to be able to call proj4.def directly on an object return bu proj4.defs...

Any opinion on these ?

ahocevar:
A proj string does not contain an identifier, a wkt string might contain one. Either way, the `proj4.defs(id, def)` function is only needed when you're relying on identifiers. If you just want to transform something unknown, but have a def, then simply do
```js
const wgs84 = proj4(def).forward(coordinates);
// and
const projected = proj4(def).inverse(wgs84);
```
If you want to get an identifier that might maybe be contained in a wkt def, then I think this should be solved in wkt-parser, not here.

PR #482, opened 2024-10-25
Add WGS84 / UTM zones as predefined CRS

ahocevar:
For convenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ instead of writing out the whole proj definition, by default proj4 has the follo
     - 'GOOGLE'
     - 'EPSG:900913'
     - 'EPSG:102113'
+- EPSG:32601 to EPSG:32660 (WGS84 / UTM zones 1 to 60 North)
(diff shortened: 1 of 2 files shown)


christophfriedrich:
The readme doesn't reflect the code correctly: The loop adds 32600 to 32660 and 32700 to 32760, but the text only lists from 32x0**1**, and both bullet points are identical where it should be x=6 in the first and x=7 in the second point.

ahocevar:
Thanks @christophfriedrich, that's fixed already in the latest release.

christophfriedrich:
Oh ok, I didn't see the later commit. That fixes the identical lines. The codes ending in `00` are still missing though (the loop starts at zero, so adds 32600 and 32700, but they are not mentioned in the text).

ahocevar:
@christophfriedrich Good point, that's a mistake in the code then. The UTM zones start at 32601 and 32701. I'll create a follow-up pull request.

PR #490, opened 2024-11-26
Make NAD grid error columns optional

oshawa-connection:
Hello,

NTv2 files tend to be quite big for web and mobile use cases, so this change gives users the ability to specify if those columns are not included in the file. When the columns are removed we can almost half the size of the file transferred.

Currently, the NTv2 file parser requires the error fields, whereas [ESRI say that they are optional](https://github.com/Esri/ntv2-file-routines?tab=readme-ov-file#gsa-file-syntax).

I should mention that this is a bit of a first draft - I need to clean up the tests I "added", if you think that this PR is a good idea :)

Lastly, I made it so includeErrorFields argument to nadgrid is optional, assumed to be true for backwards compatibility - I don't know how to indicate this to users.


ahocevar:
Thanks, @oshawa-connection. Is there any prior art for `+nadgrids=@ignorable`? Or would it be possible to auto-detect whether error fields are missing.

oshawa-connection:
Hi @ahocevar sorry for the slowness, free time is at a premium atm :(

I took a look into auto detecting the error columns. In terms of prior art:
- GDAL and QGIS (which might use GDAL's driver, I'm not sure) require all 4 bands and error out if they aren't all there
- ESRI's file routines (https://github.com/Esri/ntv2-file-routines) the user has to specify whether there are error fields or not manually (by using the `-a` or `-d` flags respectively) and will produce incorrect results if you choose the wrong one.

I think auto detection would need a big rework of `nadgrid.js` parser, which I'm quite reluctant to do! We'd need to do some kind of "look ahead": we'd need to scan through the file, detect all of the subgrid headers to count the number of rows expected, and from that, the total size of the file, and from that, the number of bytes per subgrid record. 

What do you think? I guess you are probably looking to preserve the original function signature?

test/test.js
```diff
@@ -558,6 +558,83 @@ function startTests(chai, proj4, testPoints) {
...
+        [-44.5, 40.5, -44.498553, 40.500632], // inside the first square
+        [-60, 50, -59.999192, 50.000058], // a general point towards the middle of the grid
+        [0, 0, 0, 0] // fall back to null
+      ];
+
+      var converter;
+
+      function initializeNadgrid(buffer) {
+        proj4.nadgrid('ntv2', buffer, false);
+        proj4.defs('ntv2_from', '+proj=longlat +ellps=clrk66 +nadgrids=@ignorable,ntv2,null');
```
ahocevar, line 576:
```suggestion
        proj4.defs('ntv2_from', '+proj=longlat +ellps=clrk66 +nadgrids=@ntv2,null');
```

lib/nadgrid.js
```diff
@@ -10,11 +10,11 @@ var loadedNadgrids = {};
  * Load a binary NTv2 file (.gsb) to a key that can be used in a proj string like +nadgrids=<key>. Pass the NTv2 file
  * as an ArrayBuffer.
  */
-export default function nadgrid(key, data) {
+export default function nadgrid(key, data, includeErrorFields) {
```
ahocevar, line 13: To make this safe to future upgrades, I'd change the signature to `nadgrid(key, data, options)`, with `options` currently only taking an `includesErrorFields` boolean option, with `true` being the default.

oshawa-connection:
Thanks for the suggestions, those are done. Should I try to reduce duplication between the `Nadgrids ntv2` tests and the `Nadgrids ntv2 no error columns` tests or is it ok like this?

ahocevar:
That’s ok, thanks! The only thing missing now is updating the docs in README.md.

oshawa-connection:
Good point, I've added that!

ahocevar (approved):
Thanks, @oshawa-connection

Issue #495, opened 2025-03-25
loading tif file as nadgrid


eladlavi:
hello. I use pro4js and working in Florida under EPSG 2236.
I copied the string from https://epsg.io/2236 for proj4js:
proj4.defs("EPSG:2236","+proj=tmerc +lat_0=24.3333333333333 +lon_0=-81 +k=0.999941177 +x_0=200000.0001016 +y_0=0 +ellps=GRS80 +nadgrids=us_noaa_FL.tif +units=us-ft +no_defs +type=crs");

I did it many times in the past for other projections and everything always worked like a charm,  but this is the first time I encounter: "+nadgrids=us_noaa_FL.tif".
So when trying to convert (from WGS84 for instance) I get this error: "Unable to find mandatory grid 'us_noaa_FL.tif'"
So I removed this part but then the conversion was off by few inches (for my application it is critical).
So I tried this:
I send an http request from my JavaScript code (browser based, not server side) to get the file (XMLHttpRequest and I read the response as arraybuffer (by referring to "response" and not "responseText". I made sure that the length of the array is identical to the number of bytes of the file.
Here's the file by the way:
https://cdn.proj.org/us_noaa_FL.tif
then I do this with the response:
proj4.nadgrid('key', response);
[I got this idea from here: http://proj4js.org/ under "Grid Based Datum Adjustments" where they use a gsb file]
but then I get this error:
Uncaught RangeError: Offset is outside the bounds of the DataView
    at DataView.prototype.getFloat32 (<anonymous>)

I am assuming that it is something with reading the file in a binary manner but I am not sure how to feed proj4 the binary data, meaning the content of the file. The exact type of the response is: ArrayBuffer(8374)    (again, I didn't use responseText of XMLHttpRequest but response instead).
This 8374 is exactly the size of the file.

Please help!
Thank you.

ahocevar:
Proj4js only supports ntv2 files for nadgrids. You can use this tool to convert tif to ntv2: https://github.com/OSGeo/PROJ-data/blob/85e7fca0e2912e566cfec630c682dee1b7be2e3d/grid_tools/ntv2_to_gtiff.py.

ahocevar:
I think the way forward will be to add support for tif gridshifts, maybe by leveraging geotiff.js. A pull request would be appreciated. Do you think you'd be able to work on this?

eladlavi:
Firstly, thank you very much for the help.
Secondly, is there a converter tool that can work instead ?

[...]

PR #503, opened 2025-05-14
Geotiff grid shift support 

oshawa-connection:
As discussed in #495 I have created a first draft to see if the approach is ok. 
Users have to do something like this in their code. Not particularly friendly, but it does match roughly how users have to load NTV2 grids from an arraybuffer. 

```javascript
proj4.defs("EPSG:27700_tif","+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +ellps=airy +nadgrids=britishtif +units=m +no_defs +type=crs");

async function loadGeotiff() {
    const response = await fetch('uk_os_OSTN15_NTv2_OSGBtoETRS.tif');
    const arrayBuffer = await response.arrayBuffer();
    const tiff = await fromArrayBuffer(arrayBuffer);
    const image = await tiff.getImage();
    const extent = image.getBoundingBox(); // comes out as: [-9, 48.99166666666667, 2.0166666666666657, 61]
    const rasters = await image.readRasters();
    
    const fakeExtent = [-9,49, 2, 61]; // ATTENTION! Need this part to match what comes from equivalent ntv2 file.
    const result = proj4.nadtifgrid('britishtif', rasters, {imageWidth:image.getWidth(), imageHeight:image.getHeight() extent: fakeExtent });
    
    console.log(proj4('EPSG:4326', 'EPSG:27700_tif', wgs84Coords)); // matches
}
```

~~Secondly, I am a little bit stuck because of a small (but significant) difference between the british `.gsb` and `.tif` calculated extents - I'm not sure how to handle this. When proj4js reads from the british NTV2 file (OSTN15_NTv2_OSGBtoETRS.gsb), it determines the extent to be [-9,49, 2, 61]; whereas geotiff.js returns [-9, 48.99166666666667, 2.0166666666666657, 61] from uk_os_OSTN15_NTv2_OSGBtoETRS.tif which results in differences in the calculated projected coordinate.~~

~~In QGIS, there is no difference between these two files, so I'm not sure what the cause of this difference is....~~


ahocevar:
Thanks for putting this together, @oshawa-connection. The general approach looks good!

Like you suggest, the API could be simplified - since the configuration object contains geotiff.js specific data anyway, you could pass the geotiff object directly. Instead of introducing a new `proj4.nadtifgrid` function, the existing `proj4.nadgrid` could be used and detect a TIFF if the provided `data` is not an instance of `ArrayBuffer`.
```js
proj4.defs("EPSG:27700_tif","+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +ellps=airy +nadgrids=britishtif +units=m +no_defs +type=crs");

async function loadGeotiff() {
    const response = await fetch('uk_os_OSTN15_NTv2_OSGBtoETRS.tif');
    const arrayBuffer = await response.arrayBuffer();
    const tiff = await fromArrayBuffer(arrayBuffer);
    
    await proj4.nadgrid('britishtif', tiff).ready;
    
    console.log(proj4('EPSG:4326', 'EPSG:27700_tif', wgs84Coords)); // matches
}

oshawa-connection:
Thanks for the feedback, that's implemented so hopefully a bit easier on users now :) 
- I found a good example of a GeoTIFF with subgrids, those are handled
- I tested these changes against a handful of different .tif files + coord systems, but in my automated tests I only used the Quebec one (which has subgrids) as I don't want to add too many big files to the project. If you want more tests I can potentially clip the tiffs to make them smaller.
- Sadly, I had to add a dev dependency on geotiff.js library to let me run tests

ahocevar (approved):
Thanks @oshawa-connection, this is a great addition!

PR #523, opened 2025-07-24
readme: update TypeScript instructions

mattico:
Since `@types/proj4` isn't needed anymore and is deprecated, update the readme to match.


ahocevar (approved):
Thanks, @mattico

PR #527, opened 2025-08-15
Add EPSG 5041, 5042 to predefined

rotu:
These complement EPSG 32601...32660 and 32701...32770, which cover the rest of the world.


rotu:
I'm not sure why the order is "northing, easting" as contrasted with UTM.

It seems for EPSG:32661, EPSG:32761, this should also have the parameter `+axis=ned`.
And EPSG:5041, EPSG:5042 are the same with the default axes (`+axis=enu`).
But `enforceAxis` is off by default, so the user will probably get 5041/5042 in practice.

I'm going to switch over to 5041/5042, which seems less likely to cause confusion.

ahocevar (approved):
Thanks, @rotu

PR #532, opened 2025-11-11
Fix projjson when used with defs(code, projjson)

ahocevar:
When registering a projection code with a projjson definition using `proj4.defs(code, projjson)`, the projjson object is never converted into a projection definition.

This pull request fixes that.

### DIFF
--- a/lib/defs.js
+++ b/lib/defs.js
@@ -73,6 +73,9 @@ function defs(name) {
       } else {
         defs[/** @type {string} */ (name)] = wkt(arguments[1]);
       }
+    } else if (typeof def === 'object' && !('projName' in def)) {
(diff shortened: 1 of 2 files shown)


lib/defs.js
```diff
@@ -73,6 +73,9 @@ function defs(name) {
       } else {
         defs[/** @type {string} */ (name)] = wkt(arguments[1]);
       }
+    } else if (typeof def === 'object' && !('projName' in def)) {
```
but0n: I think this will cause an error if `def` is `null` or `undefined` as `typeof null = 'object'`.
```
TypeError: def is not an Object. (evaluating '"projName" in def')
```

PR #535, opened 2025-11-17
fix(ob_tran): handle o_proj=longlat forward/inverse properly

### LINKED ISSUE #536: ob_tran projection does not handle o_proj=longlat properly
The new `ob_tran` projection does not work as expected when going forward or backwards from `longlat` / `identity`. This is because of the conversion between degrees and radians at the boundary. A special case is required to undo that automatic boundary conversion if the inner projection is `longlat`/`identity`.

potion-cellar:
Fixes #536 and adds test cases for this.

### DIFF
--- a/test/testData.js
+++ b/test/testData.js
@@ -745,50 +745,7 @@ var testPoints = [
     }
   },
   {
-    code: `GEOGCRS["unnamed",
-    BASEGEOGCRS["unknown",
-        DATUM["unknown",
-            ELLIPSOID["unknown",6371229,0,
-                LENGTHUNIT["metre",1,
-                    ID["EPSG",9001]]]],
-        PRIMEM["Greenwich",0,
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @potion-cellar

test/testData.js
```diff
@@ -726,6 +726,42 @@ var testPoints = [
       xy: -4
     }
   },
+  {
+    code: '+proj=ob_tran +o_proj=eqearth +o_lat_p=85 +o_lon_p=10',
+    ll: [20, 11],
+    xy: [2841069.7339, 808313.2811],
+    acc: {
+      ll: 1,
+      xy: -4
```
ahocevar, line 735: Why did I approve this? Note to self: the accuracy here allows pretty much any value.

ahocevar: Anyway, ob_tran is fixed with #573.

PR #573, opened 2026-04-17
Fix ob_tran forward/inverse asymmetry with rotated-pole CRS

### LINKED ISSUE #572: ob_tran forward/inverse asymmetry with rotated-pole CRS (v2.20.8)
Just a disclaimer: I'm mostly a Python guy, and definitely not really a geospatial/map projections person. Lots of this investigation has been LLM aided, but I've double checked the robots' working carefully and as best as I can tell, it's identified a real issue.

TLDR; If I try to transform and then inverse transform a lat-lon pair with this projection/CRS, I wind up off by about 64.74 degrees. This is about 2* (360- offset), which makes me think there might be a missing minus sign somewhere?

At the end of this issue, I'll link a repo that reproduces the bug as reliably as I can. I'm going to have a crack at solving it, but I've never used javascript outside the browser before, so this is all very new to me... 

```
const SWWA =
  "+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs";
```

N.B:  Everything below was written by ChatGPT, after getting it to fight it out with Claude to disprove each others hypotheses about the issue. I've tried to force it to be as clear and concise as possible, and independently verified everything it claims.

## Summary

For the rotated-pole CRS below, `proj4js` v2.20.8 produces a forward transform (`EPSG:4326` → rotated CRS) that is not consistent with its inverse transform.

A simple round-trip test fails by ~64.74° in longitude:

* `forward([116, -32])` → `[102.10, -30.46]`
* `inverse(forward([116, -32]))` → `[51.26, -32.00]`  ❌ expected ≈ `[116, -32]`

This reproduces outside application code and appears to differ from `pyproj` / PROJ 9 for the same CRS string.

## Environment

| Package | Version      |
| ------- | ------------ |
| proj4js | 2.20.8       |
| Node.js | 23.11.0      |
| pyproj  | 3.x / PROJ 9 |

## CRS

Rotated-pole CRS derived from CF `rotated_pole` metadata:

```proj
+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs
```

CF metadata used:

* `grid_north_pole_latitude = 60.31` → `o_lat_p = 60.31`
* `grid_north_pole_longitude = 147.63`

Using the standard pyproj CF mapping:

* `lon_0 = grid_north_pole_longitude + 180 = 327.63`

## Minimal JS repro

```js
import proj4 from "proj4";

const SWWA =
  "+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs";

const p = proj4("EPSG:4326", SWWA);

const points = [
  [116, -32],
  [108, -25],
  [122, -37],
];

for (const pt of points) {
  const fwd = p.forward(pt);
  const rt = p.inverse(fwd);

  console.log(
    "input:", pt,
    "forward:", fwd.map((v) => +v.toFixed(4)),
    "round-trip:", rt.map((v) => +v.toFixed(4))
  );
}
```

### proj4js output

```text
input: [116, -32] → forward: [102.1034, -30.4611] → round-trip: [51.26, -32]
input: [108, -25] → forward: [90.9087, -28.5904] → round-trip: [43.26, -25]
input: [122, -37] → forward: [110.1854, -31.6926] → round-trip: [57.26, -37]
```

The latitude round-trips correctly, but longitude is offset by a large, systematic amount.

## Python / pyproj comparison

Using the same proj string in pyproj:

```python
from pyproj import Transformer

crs = (
    "+proj=ob_tran +o_proj=longlat +o_lon_p=0 "
    "+o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs"
)

t = Transformer.from_crs("EPSG:4326", crs, always_xy=True)
inv = Transformer.from_crs(crs, "EPSG:4326", always_xy=True)

pt = (116, -32)

fwd = t.transform(*pt)
rt = inv.transform(*fwd)

print("forward:", fwd)
print("round-trip:", rt)
```

### pyproj output

```text
forward: (153.4417, -5.8948)
round-trip: (116.0000, -32.0000)
```

pyproj / PROJ produces a consistent round-trip and a materially different forward result.

## Notes

* This is not due to misuse of the proj4js API:

  * `proj4("EPSG:4326", SWWA, point)` and `p.forward(point)` give the same result.
  * `proj4(SWWA, "EPSG:4326", point)` and `p.inverse(point)` give the same result.
* The issue reproduces without any map or rendering library involved.

## Hypothesis (unconfirmed)

The longitude error is consistently ~64.74°, which is approximately `2 × 32.37°`, where:

* `lon_0 = 327.63° ≡ -32.37° (mod 360)`

This suggests there may be an issue in how `lon_0` is applied in the forward path for this `ob_tran + o_proj=longlat` case.

This is only a hypothesis based on the observed error pattern; the repro above is the main issue.

## Related

* proj4js issue #536 (`ob_tran` + `o_proj=longlat` handling)
* fix reportedly shipped in v2.20.2

___

Demo repo with reproducible example: https://github.com/charles-turner-1/proj4js-crs-bug

charles-turner-1:
Closes #572.

Full disclosure: This was written by codex.

### DIFF
--- a/test/testData.js
+++ b/test/testData.js
@@ -792,6 +792,15 @@ var testPoints = [
     }
   },
   {
+    code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs',
+    ll: [116, -32],
+    xy: [153.4417, -5.8948],
+    acc: {
+      ll: 1,
+      xy: -4
+    }
+  },
+  {
     code: '+proj=ob_tran +o_proj=moll +o_alpha=5 +o_lon_c=40 +o_lat_c=-10',
     ll: [10, 5],
     xy: [-154995.9625, -8241537.7451],
--- a/test/testData.js
+++ b/test/testData.js
@@ -740,119 +740,63 @@ var testPoints = [
   {
     code: '+proj=ob_tran +o_proj=moll +o_lat_p=45 +o_lon_p=-90',
     ll: [-2, -1],
-    xy: [-7421459.0847, -5444548.62239],
-    acc: {
-      ll: 5,
-      xy: 0
-    }
+    xy: [-7421459.08469763, -5444548.62238893]
   },
   {
     code: '+proj=ob_tran +o_proj=eqearth +o_lat_p=85 +o_lon_p=10',
     ll: [20, 11],
-    xy: [2841069.7339, 808313.2811],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
-  },
-  {
-    code: '+proj=ob_tran +o_proj=eqearth +o_lat_p=85 +o_lon_p=10',
-    ll: [20, 11],
-    xy: [2841069.7339, 808313.2811],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
+    xy: [2841069.73392768, 808313.281062982]
   },
   {
     code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=35',
     ll: [-105, 40],
-    xy: [-60.8425, 32.0797],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
+    xy: [-60.8425058899586, 32.0797099050498]
+
   },
   {
     code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=35 +lon_0=-113 +R=6371229 +no_defs +type=crs',
     ll: [-105, 40],
-    xy: [6.3262, -14.6381],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
+    xy: [6.32623159167842, -14.6380639304457]
   },
   {
     code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=31.758312 +lon_0=-92.402969 +R=6371229 +no_defs +type=crs',
     ll: [-105, 40],
-    xy: [-10.0777, -17.2983],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
+    xy: [-10.0776833744144, -17.2983149595925]
   },
   {
     code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=60.31 +lon_0=327.63 +a=6371229 +no_defs',
     ll: [116, -32],
-    xy: [153.4417, -5.8948],
-    acc: {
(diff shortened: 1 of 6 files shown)


test/testData.js
```diff
@@ -773,24 +773,33 @@ var testPoints = [
...
-  {
-    code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=31.758312 +lon_0=-92.402969 +R=6371229 +no_defs +type=crs',
-    ll: [-105, 40],
-    xy: [-10.0777, -17.2983],
-    acc: {
-      ll: 1,
-      xy: -4
-    }
-  },
+  {
```
ahocevar:
```suggestion
  {
    code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=35 +lon_0=-113 +R=6371229 +no_defs +type=crs',
    ll: [-105, 40],
    xy: [6.32623159167842, -14.6380639304457],
    acc: {
      ll: 7,
      xy: 2
    }
  },
  {
    code: '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=31.758312 +lon_0=-92.402969 +R=6371229 +no_defs +type=crs',
    ll: [-105, 40],
    xy: [-10.0776833744144, -17.2983149595925],
    acc: {
      ll: 7,
      xy: 2
    }
  },
  {
```

ahocevar (changes requested):
Let's keep the existing tests!

ahocevar:
Thanks for your effort, @charles-turner-1! Your changes aside, looking at the ob_tran implementation and tests, I see that there are more issues than the one you fixed. I'm going to add a commit to your pull request.

ahocevar:
All the ob_tran tests had a tolerance that would accept almost any value. In 9a20c41, I removed these tolerances, and updated the value to those returned by `gdaltransform`. This revealed another issue: missing authalic handling in equearth, which I also added.

ahocevar (approved):

ahocevar:
Actually I forgot the `+over` handling. I just created #574 for that.

potion-cellar:
@ahocevar @charles-turner-1 Thank you both for finishing this up properly! I had just run into the original issue with a projection that used lon_0.

PR #576, opened 2026-05-15
Fix enforceAxis for z and arbitrary orders

### LINKED ISSUE #575: `forward(...,true)` does not respect vertical axis
`proj4` ignores the distinction between up and down for non-geocentric transforms, instead preserving the third coordinate as-is. This is the case even when specifying `enforceAxis`:

```js
import assert from 'node:assert/strict';
import proj4 from 'proj4';

const enu = '+proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees +axis=enu';
const ned = '+proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees +axis=ned';

const arrayInput = [10, 20, 30];
const arrayExpected = [20, 10, -30];
const arrayActual = proj4(enu, ned).forward(arrayInput, true);

console.log('Array input:', arrayInput);
console.log('Array actual:', arrayActual);
console.log('Array expected:', arrayExpected);

assert.deepEqual(
   arrayActual,
   arrayExpected,
   [
      'proj4js did not negate Z when converting from +axis=enu to +axis=ned.',
      `Array input: ${JSON.stringify(arrayInput)}`,
      `Array actual: ${JSON.stringify(arrayActual)}`,
      `Array expected: ${JSON.stringify(arrayExpected)}`,
   ].join('\n'),
);

console.log('Repro did not fail; the bug may have been fixed.');
```

ahocevar:
Fixes #575.

This pull request also simplifies the `adjust_axis()` function and avoids repeated checks for point object/array in `transform()`.

### DIFF
--- a/lib/Proj.js
+++ b/lib/Proj.js
@@ -32,6 +32,8 @@ function Projection(srsCode, callback) {
   this.init = null;
   /** @type {string} */
   this.name;
+  /** @type {string} */
+  this.axis;
   /** @type {Array<string>} */
   this.names = null;
   /** @type {string} */
--- a/lib/adjust_axis.js
+++ b/lib/adjust_axis.js
@@ -1,57 +1,79 @@
-export default function (crs, denorm, point) {
-  var xin = point.x,
-    yin = point.y,
-    zin = point.z || 0.0;
-  var v, t, i;
+var order = ['x', 'y', 'z'];
+
+/**
+ * Convert a point in a given CRS axis order to ENU (east/north/up) order
+ * @param {import('./defs').ProjectionDefinition} crs
+ * @param {import('./core').InterfaceCoordinates} point
+ * @returns {import('./core').InterfaceCoordinates | null}
+ */
+export function adjustAxisToEnu(crs, point) {
   /** @type {import("./core").InterfaceCoordinates} */
-  var out = {};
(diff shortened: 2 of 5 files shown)


rotu:
This is a ton better!

Nit: I would *not* move `var` statements from the top of the `adjust_axis.js` function since it's misleading --- `var` is always scoped to the enclosing function. Either use `let`/`const` or keep `var` declarations at the top.

I think there's still an issue with the axis order:
```js
require('.')(
 '+proj=longlat +axis=neu',
 '+proj=longlat +axis=uen').forward([10,20,30],true);
// [ 10, 30, 20 ]
// expected: [30, 20, 10]

require('.')(
 '+proj=longlat +axis=uen',
 '+proj=longlat +axis=uen').forward([10,20,30],true);
// [ 29.999999999999996, 10, 20 ]
// expected: [10, 20, 30] (approximately)
```

ahocevar:
Both fixed, @rotu. I don't think that the arbitrary order example you provided has relevance in practice, but now we can also handle that and it's a good test case.

rotu:
You're right that the example is unrealistic - I chose it because it's obviously wrong and easy to troubleshoot.

The code looks good to me!

rotu (approved):
lgtm!

PR #579, opened 2026-06-22
Add +lon_wrap support

### LINKED ISSUE #180: Add +lon_wrap support
I need to convert coordinates into lon/lat wrapped to the 0,360 longitude range. This can be done with `+proj=longlat +ellps=WGS84 +datum=WGS84 +lon_wrap=180 +no_defs`. See the [proj.4 docs](https://trac.osgeo.org/proj/wiki/GenParms#lon_wrapover-LongitudeWrapping) on that. It seems lon_wrap is not supported by proj4js. Is there a chance that this feature could be added? It would be very useful in dealing with certain scientific datasets.

cycsmail:
Fixes #180

Adds parsing of the `+lon_wrap` parameter and wraps longitude output into the range centered on that value, matching PROJ (results land in `[lon_wrap - 180, lon_wrap + 180]`). So `+lon_wrap=180` puts output in 0..360 and -170 comes out as 190, while leaving the default -180..180 behavior untouched when it isn't set. Added a test covering the wrapped and unwrapped cases.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 ## Change log
 
-- 2.20.9: Add Eckert VI (eck6) projection; add geotiff@3 support (geotiff is now an optional peer dependency); fix unknown datum handling to match PROJ behavior; fix gamma without alpha in omerc; fix ob_tran lam offset and +over handling; fix enforceAxis for z and arbitrary axis orders
+- 2.20.9: Add Eckert VI (eck6) projection; add geotiff@3 support (geotiff is now an optional peer dependency); add support for `+lon_wrap`; fix unknown datum handling to match PROJ behavior; fix gamma without alpha in omerc; fix ob_tran lam offset and +over handling; fix enforceAxis for z and arbitrary axis orders
 
 - 2.20.8: Add back TypeScript type definitions to the npm package (accidentally removed in 2.20.7)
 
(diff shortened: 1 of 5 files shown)


ahocevar (changes requested):
Thanks for your contribution, @cycsmail. The code looks good, I only have one minor concern:

There is an Infinite-loop edge case: a non-finite `point.x` (`Infinity`) would spin the `while` forever. An out-of-range/garbage longitude reaching this point is unlikely, but a bounded form (reuse `adjust_lon`) would be more robust than open while loops.

cycsmail:
good point, swapped the while loops for adjust_lon on the offset, so it's a single bounded step and a non-finite x can't spin. full suite passes.

changelog.md
```diff
@@ -1,6 +1,6 @@
 ## Change log
 
-- 2.20.9: Add Eckert VI (eck6) projection; add geotiff@3 support (geotiff is now an optional peer dependency); fix unknown datum handling to match PROJ behavior; fix gamma without alpha in omerc; fix ob_tran lam offset and +over handling; fix enforceAxis for z and arbitrary axis orders
+- 2.20.9: Add Eckert VI (eck6) projection; add geotiff@3 support (geotiff is now an optional peer dependency); add support for `+lon_wrap`; fix unknown datum handling to match PROJ behavior; fix gamma without alpha in omerc; fix ob_tran lam offset and +over handling; fix enforceAxis for z and arbitrary axis orders
```
ahocevar:
```suggestion
- 2.21.0: Add support for `+lon_wrap`
- 2.20.9: Add Eckert VI (eck6) projection; add geotiff@3 support (geotiff is now an optional peer dependency); fix unknown datum handling to match PROJ behavior; fix gamma without alpha in omerc; fix ob_tran lam offset and +over handling; fix enforceAxis for z and arbitrary axis orders
```

ahocevar (approved):
Thanks, @cycsmail

PR #580, opened 2026-06-29
Default lon_0 to 0 in aeqd, merc and lcc

mahirhir:
#560 added `this.long0 = this.long0 || 0` to sinu, and #578 added the `x0`/`y0` defaults to aeqd. These three projections default `x0`/`y0` in `init()` but never default `long0`, so they return `NaN` when a definition omits `+lon_0`. PROJ treats a missing `lon_0` as `0`, and the other projections that already guard it (sinu, eqc, equi, ortho, qsc, robin, stere) use the same `|| 0` idiom.

`forward`/`inverse` read `this.long0` directly:

- aeqd: `adjust_lon(lon - this.long0, this.over)`
- merc: `adjust_lon(lon - this.long0, this.over)`
- lcc: `this.ns * adjust_lon(lon - this.long0, this.over)`

With `long0` undefined that subtraction is `NaN`.

Before:

```js
proj4('+proj=merc +a=6378137 +b=6378137 +units=m +no_defs', [10, 45])
// [NaN, 5621521.486192066]
proj4('+proj=lcc +lat_1=33 +lat_2=45 +lat_0=39 +datum=WGS84 +units=m +no_defs', [-100, 40])
// [NaN, NaN]
proj4('+proj=aeqd +lat_0=51.5 +datum=WGS84 +units=m +no_defs', [2, 48])
// [NaN, NaN]
```

After (identical to adding `+lon_0=0`):

```js
// merc -> [1113194.90793274, 5621521.48619207]
// lcc  -> [-6880442.71344780, 4330863.47097279]
// aeqd -> [149325.62485313, -387277.84141508]
```

Added one test point per projection in `testData.js` with `+lon_0` omitted. They fail with `NaN` before the change and round-trip through the inverse after it.

### DIFF
--- a/lib/projections/aeqd.js
+++ b/lib/projections/aeqd.js
@@ -24,6 +24,7 @@ export function init() {
   this.cos_p12 = Math.cos(this.lat0);
   this.x0 = this.x0 || 0;
   this.y0 = this.y0 || 0;
+  this.long0 = this.long0 || 0;
   // flattening for ellipsoid
   this.f = this.es / (1 + Math.sqrt(1 - this.es));
 }
--- a/lib/projections/lcc.js
+++ b/lib/projections/lcc.js
@@ -35,6 +35,7 @@ export function init() {
   }
   this.x0 = this.x0 || 0;
   this.y0 = this.y0 || 0;
+  this.long0 = this.long0 || 0;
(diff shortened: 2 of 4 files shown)


ahocevar (approved):
Thanks, @greymoth-jp

PR #582, opened 2026-06-29
fix(nzmg): add iterations to default export

### LINKED ISSUE #581: NZMG inverse: ~17 m drift on EPSG:27200 round-trip
hello! I believe there is a bug in the NZMG projection. I am seeing a horizontal drift of ~17m for the roundtrip `forward(inverse(x))` using EPSG:27200 (NZGD49 / New Zealand Map Grid). I was able to reproduce on `proj4@2.20.9`. Seems like this may have been broken since `nzmg.js` was migrated from CommonJS to ESM in 02d0f51

```js
const proj4 = require('proj4');
const NZMG = '+proj=nzmg +lat_0=-41 +lon_0=173 +x_0=2510000 +y_0=6023150 +ellps=intl +no_defs';
const input = [2152713.585, 5442524.565];
const out = proj4(NZMG).forward(proj4(NZMG).inverse(input));
// out = [2152730.0826, 5442518.4057]    // ~17.6 m off
// expected ≈ input                       // sub-mm
```

For comparison, GDAL returns the input unchanged:

```
$ echo "2152713.585 5442524.565" | gdaltransform -s_srs EPSG:27200 -t_srs EPSG:27200
2152713.585 5442524.565 0
```

`proj4@2.3.15` also returns the input to sub-mm.

I believe the cause is that `iterations` is missing from the default export. So `extend(this, ourProj)` in `Proj.js` can't copy it onto the instance, `this.iterations` is undefined, and the loop runs 0 times.. I see there was a follow-up commit da05fc0 on the same day to add the missing default-export entries to `laea.js` and `stere.js`. So the fix should be simple for `nzmg.js` too? 

```diff
 export default {
   init: init,
   forward: forward,
   inverse: inverse,
-  names: names
+  names: names,
+  iterations: iterations
 };
```

jedabanat:
Fixes #581  & adds a test case

`iterations` was missing from the default export for when `nzmg.js` so `extend(this, ourProj)` in `Proj.js` never copies it onto the instance and the inverse fixed-point loop runs 0 times. This caused a ~17 m drift on `forward(inverse(x))` round-trips for EPSG:27200.

### DIFF
--- a/lib/projections/nzmg.js
+++ b/lib/projections/nzmg.js
@@ -222,5 +222,6 @@ export default {
   init: init,
   forward: forward,
   inverse: inverse,
-  names: names
+  names: names,
+  iterations: iterations
 };
--- a/test/proj4.test.mjs
+++ b/test/proj4.test.mjs
@@ -873,4 +873,15 @@ describe('proj4', function () {
       }, 'Invalid value for o_lat_1: o_lat_1 should be different from zero', 'should work');
     });
   });
+
+  describe('nzmg projection', function () {
+    var proj = proj4('+proj=nzmg +lat_0=-41 +lon_0=173 +x_0=2510000 +y_0=6023150 +ellps=intl +no_defs');
+    var nzmg = [2152713.585, 5442524.565];
+    it('should round trip through inverse and forward', function () {
+      var lonlat = proj.inverse(nzmg);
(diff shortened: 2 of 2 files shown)


ahocevar (approved):
Thanks, @jedabanat

PR #584, opened 2026-07-02
Fix Van der Grinten forward returning NaN on the equator and central meridian

spokodev:
The Van der Grinten projection (`+proj=vandg`) forward transform returns `[NaN, NaN]` for every point on the equator and on the central meridian (and at the origin and poles).

```js
const vandg = proj4('+proj=vandg +R=6371000 +no_defs');
vandg.forward([0, 45]);  // [NaN, NaN], expected [0, 5363026.34]
vandg.forward([45, 0]);  // [NaN, NaN], expected [5003771.70, 0]
```

Off-axis points are fine, which is why it was not caught earlier.

## Cause

In `lib/projections/vandg.js` `forward`, the equator branch (`|lat| <= EPSLN`) and the central-meridian/pole branch compute the correct `x`/`y` but do not return; execution falls through to the general formulas. Those divide by `dlon` (zero on the central meridian) and by `sinth + costh - 1` (zero on the equator), so the result becomes `NaN`. The central-meridian branch even carries a leftover `//  return(OK);` comment from the ported GCTP/PROJ C source, marking the return that was dropped.

## Fix

Return the point from each special-case branch, matching the file's own epilogue (`p.x = x; p.y = y; return p;`) and the intent of the ported source.

## Testing

Added `test/vandg-axis.test.mjs` with equator and central-meridian points whose expected values come from the closed-form Van der Grinten I on a sphere. Both fail on the current code (`NaN`) and pass with the fix. The full suite stays green (4246 tests).

### DIFF
--- a/test/testData.js
+++ b/test/testData.js
@@ -2610,6 +2610,16 @@ var testPoints = [
     xy: [343065.9150369169, 2254539.6570760156]
   },
   {
+    code: '+proj=vandg +R=6371000 +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs',
+    ll: [0, 45],
+    xy: [0, 5363026.34343254]
+  },
+  {
+    code: '+proj=vandg +R=6371000 +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs',
+    ll: [45, 0],
+    xy: [5003771.699005143, 0]
+  },
+  {
     code: '+proj=merc +ellps=plessis +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs',
     ll: [10, 50],
     xy: [1112913.211791464, 6413002.836941121]
--- a/lib/projections/vandg.js
+++ b/lib/projections/vandg.js
@@ -30,6 +30,9 @@ export function forward(p) {
   if (Math.abs(lat) <= EPSLN) {
     x = this.x0 + this.R * dlon;
     y = this.y0;
+    p.x = x;
+    p.y = y;
+    return p;
   }
   var theta = asinz(2 * Math.abs(lat / Math.PI));
   if ((Math.abs(dlon) <= EPSLN) || (Math.abs(Math.abs(lat) - HALF_PI) <= EPSLN)) {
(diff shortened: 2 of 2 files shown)


test/vandg-axis.test.mjs
```diff
@@ -0,0 +1,17 @@
...
+    var xy = proj4('WGS84', 'vandgSphere').forward([0, 45]);
+    assert.closeTo(xy[0], 0, 0.001, 'x is close');
+    assert.closeTo(xy[1], 5363026.34343254, 0.001, 'y is close');
+  });
+  it('should project a point on the equator', function () {
+    var xy = proj4('WGS84', 'vandgSphere').forward([45, 0]);
+    assert.closeTo(xy[0], 5003771.699005143, 0.001, 'x is close');
+    assert.closeTo(xy[1], 0, 0.001, 'y is close');
+  });
+});
```
ahocevar: Just add two cases to test/testData.js, instead of creating a separate test file.

ahocevar (changes requested):
Thanks for your contribution, @spokodev. The change looks good, but please address my comment about the added tests.

spokodev:
Done. The two cases now live in test/testData.js and the separate file is gone. They cover both the central meridian and equator branches, forward and inverse, and the full suite stays green.

test/testData.js
```diff
@@ -2603,6 +2603,18 @@ var testPoints = [
     code: '+proj=lcc +lat_0=-90.0 +lon_0=81.0 +lat_1=-72.66666666666674 +lat_2=-75.3333333333334 +x_0=0.0 +y_0=0.0 +ellps=GRS80 +no_defs',
     ll: [90, -70],
     xy: [343065.9150369169, 2254539.6570760156]
+  },
+  /* vandg: a point on the central meridian (dlon = 0) must not return NaN */
```
ahocevar: No need to add a comment here. Please remove.

test/testData.js
```diff
@@ -2603,6 +2603,18 @@ var testPoints = [
...
     ll: [90, -70],
     xy: [343065.9150369169, 2254539.6570760156]
+  },
+  /* vandg: a point on the central meridian (dlon = 0) must not return NaN */
+  {
+    code: '+proj=vandg +R=6371000 +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs',
+    ll: [0, 45],
+    xy: [0, 5363026.34343254]
+  },
+  /* vandg: a point on the equator (lat = 0) must not return NaN */
```
ahocevar: Same.

ahocevar (changes requested):

spokodev:
Both comments removed.

ahocevar (approved):

PR #585, opened 2026-07-15
Fix plessis ellipsoid: 6355863 is the semi-minor axis, not the inverse flattening

jiayuasu:
The `plessis` ellipsoid entry stores `6355863` in the `rf` (inverse flattening) slot, but `6355863` is Plessis 1817's semi-minor axis **b** — the inverse flattening is ~308.64. `deriveConstants` reads it as an inverse flattening and derives a near-sphere (`b = (1 - 1/6355863) * a ≈ 6376521.997`), so any transform using `+ellps=plessis` is off.

Mercator of `(10, 50)` before the fix (near-sphere):

```js
proj4('+proj=longlat +datum=WGS84 +no_defs',
      '+proj=merc +ellps=plessis +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs').forward([10, 50]);
// [1112913.21, 6444643.06]
```

PROJ 9.5.1 / pyproj gives `[1112913.21, 6413002.84]` — a ~31.6 km difference in northing:

```python
from pyproj import Transformer
Transformer.from_crs("EPSG:4326", "+proj=merc +ellps=plessis +units=m",
                     always_xy=True).transform(10, 50)
# (1112913.21, 6413002.84)   # a=6376523, b=6355863, rf≈308.64
```

The fix changes `rf: 6355863` → `b: 6355863`, consistent with the other `b:`-defined ellipsoids already in the table. Added a regression test against PROJ's values; the full suite passes (4318 tests).

### DIFF
--- a/test/testData.js
+++ b/test/testData.js
@@ -2604,7 +2604,6 @@ var testPoints = [
     ll: [90, -70],
     xy: [343065.9150369169, 2254539.6570760156]
   },
-  // plessis: 6355863 is the semi-minor axis, not the inverse flattening (values from PROJ 9.5.1)
   {
     code: '+proj=merc +ellps=plessis +lon_0=0 +x_0=0 +y_0=0 +units=m +no_defs',
     ll: [10, 50],
--- a/test/testData.js
+++ b/test/testData.js
@@ -2603,6 +2603,12 @@ var testPoints = [
     code: '+proj=lcc +lat_0=-90.0 +lon_0=81.0 +lat_1=-72.66666666666674 +lat_2=-75.3333333333334 +x_0=0.0 +y_0=0.0 +ellps=GRS80 +no_defs',
     ll: [90, -70],
     xy: [343065.9150369169, 2254539.6570760156]
+  },
+  // plessis: 6355863 is the semi-minor axis, not the inverse flattening (values from PROJ 9.5.1)
+  {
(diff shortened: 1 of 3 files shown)


ahocevar (changes requested):
The change seems to make sense, but the test does not. Instead of the test you have now, add the case you brought up in the description to `testData.js`.

jiayuasu:
Moved the test case to testData.js and removed the separate test block. Full suite passes.

test/testData.js
```diff
@@ -2603,6 +2603,12 @@ var testPoints = [
     code: '+proj=lcc +lat_0=-90.0 +lon_0=81.0 +lat_1=-72.66666666666674 +lat_2=-75.3333333333334 +x_0=0.0 +y_0=0.0 +ellps=GRS80 +no_defs',
     ll: [90, -70],
     xy: [343065.9150369169, 2254539.6570760156]
+  },
+  // plessis: 6355863 is the semi-minor axis, not the inverse flattening (values from PROJ 9.5.1)
```
ahocevar:
```suggestion
```

ahocevar (changes requested):
~~The Ellipsoid.js file is auto-generated using scripts/updateEllipsoids.mjs. So the change here will not persist. The right fix is to fix the script.~~

Actually this change is one of the manual entries in that file, so we're good.

ahocevar (approved):

PR #595, opened 2026-07-26
Fix ortho x0

JScones:
Apply the false easting `this.x0` on x in ortho projections.

This makes it consistent with y0, and the inverse already subtracts both. Now round-trips correctly.


ahocevar (changes requested):
Instead of adding a full test case, add a fixture in `test/testData.js`. Thanks in advance.

ahocevar (approved):
Thanks, @JScones

PR #598, opened 2026-08-03
Preserve z when enforceAxis=true for omerc and robin

mutaflux:
omerc and robin projections' `forward()` and `inverse()` construct a new object instead of modifying the input as other projections, so when `enforceAxis=true` the z coordinate is skipped at all by `core.js#transformer()`.

This change reapplies the z coordinate to the result of `.forward()` and `.inverse()` calls if it was present in the input of the call.

### DIFF
--- a/test/proj4.test.mjs
+++ b/test/proj4.test.mjs
@@ -357,53 +357,70 @@ describe('proj4', function () {
     assert.closeTo(r2[2], 30, 1e-6, 'uen→uen: round-trip preserves third');
   });
 
-  it('should preserve z for projections that return new objects (omerc, robin) when enforceAxis', function () {
+  it('should preserve z for omerc when enforceAxis', function () {
     var wgs = '+proj=longlat +datum=WGS84 +axis=enu';
-    var ned_omerc = '+proj=omerc +lat_0=4 +lonc=115 +alpha=53 +k=0.99984 +x_0=590476.87 +y_0=442857.65 +gamma=53.13010235415598 +ellps=WGS84 +units=m +axis=ned';
-    var ned_robin = '+proj=robin +datum=WGS84 +units=m +axis=ned';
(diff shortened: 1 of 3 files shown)


samin-propeller (approved):

test/proj4.test.mjs
```diff
@@ -357,6 +357,55 @@ describe('proj4', function () {
...
+    assert.equal(rawOmerc.forward({ x: 0.1, y: 0.3, z: 30, m: 7 }).m, 7, 'omerc raw forward: m preserved');
+    assert.equal(rawRobin.forward({ x: 0.1, y: 0.3, z: 30, m: 7 }).m, 7, 'robin raw forward: m preserved');
+    assert.equal(rawRobin.forward({ x: 0.1, y: 0.3, z: 30, m: 7 }).z, 30, 'robin raw forward: z preserved');
+
+    // ob_tran wrapping robin with a permuting axis
+    var ob_tran_robin_uen = '+proj=ob_tran +o_proj=robin +o_lat_p=45 +o_lon_p=0 +lon_0=0 +ellps=WGS84 +axis=uen';
+    var ob1 = proj4(wgs, ob_tran_robin_uen).forward({ x: 10, y: 20, z: 30 }, true);
+    assert.equal(ob1.x, 30, 'ob_tran(robin) uen: x (up) is input z');
+    assert.equal(typeof ob1.y, 'number', 'ob_tran(robin) uen: projected y present');
+    assert.equal(typeof ob1.z, 'number', 'ob_tran(robin) uen: projected z present');
```
ahocevar: A `closeTo` check would be more reliable than just checking for the presence of a number. Same with the other similar assertions above.

ahocevar (changes requested):

test/proj4.test.mjs
mutaflux: thank you for looking into the PR. I updated the tests to use value assertions.

ahocevar (approved):
Thanks, @mutaflux

PR #604, opened 2026-09-13
Route spherical tmerc to the spherical formula instead of etmerc

### LINKED ISSUE #394: +tmerc projections fail when the ellipsoid is spherical
PROJ reverts to the evenden_snyder algorithm in this case, but Proj4js throws an error stating 'incorrect elliptical usage'. 

See here for the relevant line in PROJ https://github.com/OSGeo/PROJ/blob/master/src/projections/tmerc.cpp#L597
See here for the relevant line in Proj4js https://github.com/proj4js/proj4js/blob/master/lib/projections/etmerc.js#L14

For example:

```
> require("proj4")("+proj=tmerc +a=6378137 +b=6378137 +lon_0=11.598056919260632 +x_0=1.0621731827420435e-09 +y_0=-5375403.8751633102 +k=1.0 +units=m +wktext +no_defs")
Uncaught:
Error: Incorrect elliptical usage. Try using the +approx option in the proj string, or PROJECTION["Fast_Transverse_Mercator"] in the WKT.
    at Projection.init$3 [as init] (/Users/jared/five/dev/scenario/frontend/node_modules/proj4/dist/proj4-src.js:3439:15)
    at new Projection (/Users/jared/five/dev/scenario/frontend/node_modules/proj4/dist/proj4-src.js:1649:12)
    at Projection (/Users/jared/five/dev/scenario/frontend/node_modules/proj4/dist/proj4-src.js:1595:16)
    at checkProj (/Users/jared/five/dev/scenario/frontend/node_modules/proj4/dist/proj4-src.js:2308:14)
    at proj4$1 (/Users/jared/five/dev/scenario/frontend/node_modules/proj4/dist/proj4-src.js:2312:18)
```

cpruijsen:
`+proj=tmerc` throws "Incorrect elliptical usage" when the ellipsoid is spherical.

`etmerc.js` is registered for both families. Its `names` list carries `etmerc` and
`Extended_Transverse_Mercator` alongside `tmerc`, `Transverse_Mercator` and the Gauss-Kruger
aliases, while `tmerc.js` registers only `Fast_Transverse_Mercator`. So a plain `+proj=tmerc`
string reaches `etmerc`'s guard and throws, even though `tmerc.js` already implements the exact
spherical form in its `if (!this.es)` branch.

This routes a spherical ellipsoid to that branch, reusing the delegation `+approx` already does
three lines above, and keeps the error for an explicit spherical `etmerc`.

That split matches PROJ, checked with `gdaltransform` on PROJ 9:

```
$ echo "2 1" | gdaltransform -s_srs EPSG:4326 -t_srs "+proj=tmerc +a=6400000 +b=6400000"
223413.466406322 111769.145040597 0

$ echo "2 1" | gdaltransform -s_srs EPSG:4326 -t_srs "+proj=etmerc +a=6400000 +b=6400000"
ERROR 1: PROJ: proj_create: Error 1027 (Invalid value for an argument):
etmerc: Invalid value for eccentricity: it should not be zero
```

proj4js now returns `223413.46640632232 111769.14504059685` for the first and throws for the
second. The string from #394 gives `25664.0888975901 747241.85882652` in PROJ and
`25664.088897589085 747241.8588265199` here.

Two cases added to `test/testData.js`: spherical `tmerc`, and the `tmerc` string from the issue.
Nothing changes for `es > 0`, for `+approx`, or for explicit `etmerc`.

Fixes #394


test/testData.js
```diff
@@ -2112,6 +2112,17 @@ var testPoints = [
     ll: [2, 1],
     xy: [2516532.477709202, 139083.35793371277]
   },
+  // etmerc on a sphere (es=0) is exact, see https://github.com/proj4js/proj4js/issues/394
```
ahocevar, line 2115:
```suggestion
```

test/testData.js
```diff
@@ -2112,6 +2112,17 @@ var testPoints = [
     ll: [2, 1],
     xy: [2516532.477709202, 139083.35793371277]
   },
+  // etmerc on a sphere (es=0) is exact, see https://github.com/proj4js/proj4js/issues/394
+  {
+    code: '+proj=etmerc +a=6400000 +b=6400000 +lat_1=0.5 +lat_2=2 +n=0.5',
+    ll: [2, 1],
+    xy: [223413.4664063224, 111769.14504058579]
```
ahocevar, line 2119: Trying this with cs2cs or gdaltransform, I get
```
ERROR 1: PROJ: proj_create: Error 1027 (Invalid value for an argument): etmerc: Invalid value for eccentricity: it should not be zero
ERROR 1: Translating source or target SRS failed:
+proj=etmerc +a=6400000 +b=6400000 +lat_1=0.5 +lat_2=2 +n=0.5
```
So I don't think your change in the code is the right thing to do.

ahocevar (changes requested):

test/testData.js
cpruijsen: You're right, and thanks for the `gdaltransform` output. My description was wrong: PROJ does reject spherical `etmerc`, so relaxing that guard was not the fix. I reproduced your error on PROJ 9.

The actual problem is dispatch. `etmerc.js` is registered for both families, since its `names` list carries `tmerc`, `Transverse_Mercator` and the Gauss-Kruger aliases alongside the `etmerc` ones, while `tmerc.js` registers only `Fast_Transverse_Mercator`. So a plain `+proj=tmerc` string lands in the extended guard and throws, even though `tmerc.js` already has the exact spherical form in its `if (!this.es)` branch. That is the branch PROJ uses.

Pushed b97fc86, which routes a sphere to that branch using the same delegation `+approx` already does, and keeps the error for explicit `etmerc`. Same split as PROJ:

```
$ echo "2 1" | gdaltransform -s_srs EPSG:4326 -t_srs "+proj=tmerc +a=6400000 +b=6400000"
223413.466406322 111769.145040597 0

$ echo "2 1" | gdaltransform -s_srs EPSG:4326 -t_srs "+proj=etmerc +a=6400000 +b=6400000"
ERROR 1: PROJ: proj_create: Error 1027 (Invalid value for an argument):
etmerc: Invalid value for eccentricity: it should not be zero
```

proj4js now gives `223413.46640632232 111769.14504059685` for the first and throws for the second. The #394 string gives `25664.0888975901 747241.85882652` in PROJ against `25664.088897589085 747241.8588265199` here.

I dropped the spherical `etmerc` test case, since that input should error, and kept the `tmerc` one. 4482 tests pass, lint and typecheck clean. Title and description updated, since both described the old approach.

